PR #12, opened 2020-02-17
Anastasiia.Bardyuzha: react_movies-list-filter

AnastasiiaBardyuzha:
https://AnastasiiaBardyuzha.github.io/react_movies-list-filter/

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,9 +4,23 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -4,9 +4,23 @@ import { MoviesList } from './components/MoviesList';
...
+  };
+
+  changedInput = (event) => {
+    this.setState({
+      query: event.target.value.toLowerCase(),
+    });
+  }
 
   render() {
+    const searcher = this.state.query;
```
danheim: make is easy
```suggestion
    const { query } = this.state;
```

PR #69, opened 2020-05-03
solution

cutiekate:
[DEMO LINK](https://cutiekate.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,9 +4,16 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -4,9 +4,16 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+  };
 
   render() {
+    const movieListFilter = moviesFromServer.filter(({
```
irenhh, line 12: this is not a filter, this is an array, which you get at the end of the calculations, therefore more appropriate name will be `filteredMovies` for example

PR #73, opened 2020-05-05
Solution

Qoa11a:
[Demo Link](https://qoa11a.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,9 +4,20 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -4,9 +4,20 @@ import { MoviesList } from './components/MoviesList';
...
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+  };
 
   render() {
+    const { query } = this.state;
+    const filteredMovies = [...moviesFromServer]
```
irenhh, line 13: filter copies the initial array for you so there is no need to do it manually

PR #84, opened 2020-07-02
Solution

Mishtal-Andrii:
[DEMO LINK](https://Mishtal-Andrii.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,9 +4,17 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+  };
 
   render() {
+    const { query } = this.state;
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -4,9 +4,17 @@ import { MoviesList } from './components/MoviesList';
...
+  state = {
+    query: '',
+  };
 
   render() {
+    const { query } = this.state;
+    const filteredMovies = moviesFromServer
+      .filter(({ title, description }) => (title + description)
+        .toLowerCase()
+        .includes(query.toLowerCase()));
```
danheim, line 16: make it readable
```suggestion
    const filteredMovies = moviesFromServer.filter((movie) => {
      const { title, description } = movie;
      
      return (title + description)
        .toLowerCase()
        .includes(query.toLowerCase())
    });
```

PR #114, opened 2020-09-06
solution

karinaKudatska:
[DEMO LINK](https://karinaKudatska.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,9 +4,18 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+  };
 
   render() {
+    const { query } = this.state;
+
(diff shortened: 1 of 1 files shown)


asbh25 (approved):

src/App.js
```diff
@@ -4,9 +4,18 @@ import { MoviesList } from './components/MoviesList';
...
+  state = {
+    query: '',
+  };
 
   render() {
+    const { query } = this.state;
+
+    const movies = moviesFromServer.filter(
+      movie => movie.title.toUpperCase().includes(query.toUpperCase())
+      || movie.description.toUpperCase().includes(query.toUpperCase()),
```
irenhh, line 16: usually it's converted to losercase but ok, this works as well

how about converting query at the same time as you write it in the state?

irenhh:
add debounce

irenhh (approved):

PR #121, opened 2020-09-07
Develop

JuliaNastenko:
[DEMO LINK](https://julianastenko.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,9 +4,19 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+  };
 
   render() {
+    const movies = moviesFromServer.filter(
+      movie => !this.state.query
+        || (movie.title.toLocaleLowerCase()
+          .includes(this.state.query.toLocaleLowerCase()))
+        || (movie.description.toLocaleLowerCase()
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -4,9 +4,19 @@ import { MoviesList } from './components/MoviesList';
...
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+  };
 
   render() {
+    const movies = moviesFromServer.filter(
+      movie => !this.state.query
+        || (movie.title.toLocaleLowerCase()
```
ChrisZinch, line 14: .toLocaleLowerCase()?))

ChrisZinch: Можна обійтись без "!this.state.query"

irenhh: why toLocaleLowerCase?

src/App.js
```diff
@@ -4,9 +4,19 @@ import { MoviesList } from './components/MoviesList';
...
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+  };
 
   render() {
+    const movies = moviesFromServer.filter(
+      movie => !this.state.query
```
irenhh, line 13: this is not very explicit, you better write this before filtering and do filtering according to this value

irenhh (changes requested):
use debounce

irenhh (approved):

PR #190, opened 2021-01-18
First Solution

vovador:
[DEMO LINK](https://vovador.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,12 +1,16 @@
 import React, { Component } from 'react';
 import './App.scss';
-import { MoviesList } from './components/MoviesList';
+import { MoviesList } from './components/MoviesList/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+  };
 
   render() {
+    const { query } = this.state;
+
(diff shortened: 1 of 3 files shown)


xipholena (approved):
Well done!

src/App.js
```diff
@@ -19,15 +23,21 @@ export class App extends Component {
               <div className="control">
                 <input
                   type="text"
+                  vcalue={this.state.query}
```
vsm2k17, line 26: something wrong here

src/App.js
```diff
@@ -19,15 +23,21 @@ export class App extends Component {
...
                   id="search-query"
                   className="input"
                   placeholder="Type search word"
+                  onChange={(e) => {
+                    this.setState({
+                      query: e.target.value.toLocaleLowerCase(),
```
vsm2k17, line 32: bad idea, do not limit the functionality available to the user if the user wants to write using upper case he should be able to do this, use toLowerCase behind the scene, so the user doesn't even notice those actions

vsm2k17 (changes requested):

vsm2k17 (approved):
good job

PR #201, opened 2021-01-20
Movies filter - basic task solution

VinnieJ-2k20:
- [DEMO LINK](https://VinnieJ-2k20.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,9 +4,30 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+  };
+
+  handleChange = (event) => {
+    this.setState({
+      query: event.target.value,
+    });
+  }
+
+  filterMovies = (list) => {
+    const processedList = list.filter((movie) => {
+      const searchName = this.state.query.toLowerCase();
+
+      return movie.title.toLowerCase().includes(searchName)
+        || movie.description.toLowerCase().includes(searchName);
+    });
+
+    return processedList;
+  }
 
   render() {
+    const { query } = this.state;
+
     return (
       <div className="page">
         <div className="page-content">
@@ -20,6 +41,8 @@ export class App extends Component {
                 <input
                   type="text"
                   id="search-query"
+                  value={query}
+                  onChange={this.handleChange}
                   className="input"
                   placeholder="Type search word"
                 />
@@ -27,7 +50,7 @@ export class App extends Component {
             </div>
           </div>
 
-          <MoviesList movies={moviesFromServer} />
+          <MoviesList movies={this.filterMovies(moviesFromServer)} />
         </div>
         <div className="sidebar">
           Sidebar goes here
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Movies list - Filter
 - Replace `<your_account>` with your Github username in the
- [DEMO LINK](https://<your_account>.github.io/react_movies-list-filter/)
+ [DEMO LINK](https://VinnieJ-2k20.github.io/react_movies-list-filter/)
 - Follow the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline)
 
 ## Task


AlieksieienkoAndriy (approved):

src/App.js
```diff
@@ -2,32 +2,34 @@ import React, { Component } from 'react';
...
+      filteredMovies: moviesFromServer,
+    };
+    this.applyFilteredList = this.applyFilteredList.bind(this);
+  }
+
+  applyFilteredList(filteredList) {
+    this.setState({
+      filteredMovies: filteredList,
+    });
+  }
```
varseniuk, line 20: Using constructor and 'bind' you making the solution overcomplicated.
Let's make it more simple. 
You should filter the movies according to the 'query' and then transfer filtered movies to the 'MoviesList' component. The state should be on the top level to give us the possibility to filter films by this query. 
That means that if you want to make a separate component for the 'search input', you need somehow to transfer input value from the child component('MovieFilter') to the parent component, where you will handle that input value for filtering movies. 
Understanding the topic 'lifting state up' will help you a lot.
Long story short: your 'Filter' component should receive a 'query' and 'queryChangeHandler' from the parent. Then this component should handle input changes and transfer them to the top level using lifting state up. There you will be filtering your movies according to the new query.
If you have some questions - write to your channel or visit the q&a. I will explain

src/components/MovieFilter/MovieFilter.js
```diff
@@ -0,0 +1,68 @@
...
+              type="text"
+              id="search-query"
+              value={query}
+              onChange={this.handleChange}
+              className="input"
+              placeholder="Type search word"
+            />
+          </div>
+        </div>
+      </div>
```
varseniuk, line 65: logic is correct, but let's change the structure using lifting state up (without constructor)

varseniuk (changes requested):

src/App.js
VinnieJ-2k20: @varseniuk  Okay, I`ll return to this task after learning about lifting state up.

src/App.js
```diff
@@ -1,33 +1,35 @@
...
+import { MovieFilter } from './components/MovieFilter';
+import './App.scss';
 
 export class App extends Component {
-  state = {};
+  state = {
+    movies: moviesFromServer,
+  }
+
+  queryChangeHandler = (query) => {
```
aliceblunt8, line 12: let's separate saving query to the state and filtering movies

src/App.js
```diff
@@ -1,33 +1,35 @@
...
+    const queryLowerCase = query.toLowerCase();
+
+    this.setState({
+      movies: moviesFromServer.filter(({ title, description }) => (
+        title.toLowerCase().includes(queryLowerCase)
+          || description.toLowerCase().includes(queryLowerCase)
+      )),
+    });
```
aliceblunt8, line 20: actually where's no need to save movies to the state and then update them, we can filter moviesFromServer
let's create a variable filteredMovies and move this code there, then we can pass them directly to the MovieList like this
`<MoviesList movies={filteredMovies} />`

aliceblunt8 (changes requested):

src/App.js
```diff
@@ -1,33 +1,37 @@
...
-  state = {};
+  state = {
+    query: '',
+  }
+
+  queryChangeHandler = (query) => {
+    this.setState({
+      query,
+    });
+  }
```
aliceblunt8, line 16: save value in lower case

src/components/MovieFilter/MovieFilter.js
```diff
@@ -0,0 +1,45 @@
...
+import PropTypes from 'prop-types';
+
+export class MovieFilter extends Component {
+  static propTypes = {
+    onChange: PropTypes.func.isRequired,
+  }
+
+  state = {
+    query: '',
+  }
```
aliceblunt8, line 11: you should have query in the state of ONE component (App), pass it here as prop

aliceblunt8 (changes requested):

src/App.js
VinnieJ-2k20: Fixed, but this means that all text in filter form will be in lower case. I haven't seen any search behave this way.

aliceblunt8 (approved):

PR #204, opened 2021-01-20
Completed

Serhii-Alieksieievych:
[DEMO LINK](https://Serhii-Alieksieievych.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,9 +4,17 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
+    movies: [...moviesFromServer],
+    query: '',
+  };
 
   render() {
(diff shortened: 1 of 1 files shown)


deelray (approved):

src/App.js
```diff
@@ -22,12 +30,18 @@ export class App extends Component {
                   id="search-query"
                   className="input"
                   placeholder="Type search word"
+                  value={this.state.query}
+                  onChange={(event) => {
+                    this.setState({
+                      query: event.target.value.toLowerCase(),
```
vsm2k17, line 36: bad idea, do not limit the functionality available to the user if the user wants to write using upper case he should be able  to do this

vsm2k17: use toLowerCase behind the scene, so the user doesn't even notice those actions

vsm2k17 (changes requested):
good job, almost done

vsm2k17 (approved):
good job

PR #205, opened 2021-01-20
create Input for search movie on React

KirillLutsenko:
[DEMO LINK](https://KirillLutsenko.github.io/react_movies-list-filter/)


Pasha5555 (approved):
Very nice job👍

src/App.jsx
```diff
@@ -0,0 +1,52 @@
...
+            .toLowerCase()
+            .includes(query.toLowerCase())
+        ),
+      );
+
+    return (
+      <div className="page">
+        <div className="page-content">
+          <SearchMovie
+            prevState={this.state}
```
varseniuk, line 40: You don't need to pass the whole state. Only the 'query' field

src/App.jsx
```diff
@@ -0,0 +1,52 @@
...
+    query: '',
+  };
+
+  handleChange = (event) => {
+    const { name, value } = event.target;
+
+    this.setState({
+      [name]: value,
+    });
+  }
```
varseniuk, line 18:
```suggestion
    const { value } = event.target;

    this.setState({
      query: value,
    });
  }
```
Why do you need 'name' here? 'value' would be enough

varseniuk (changes requested):

varseniuk (approved):
👍

PR #216, opened 2021-01-21
movie filter

Layrovell:
[DEMO LINK](https://Layrovell.github.io/react_movies-list-filter/)

### DIFF
--- a/src/components/SearchField/SearchField.jsx
+++ b/src/components/SearchField/SearchField.jsx
@@ -1,55 +1,27 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { MoviesList } from '../MoviesList';
 
-export class SearchField extends React.Component {
-  state = {
-    query: '',
-  };
-
-  render() {
-    const { movies } = this.props;
-
-    const searchQuery = this.state.query.toLocaleLowerCase().trim();
-    const filteredMovies = [...movies]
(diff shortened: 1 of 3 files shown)


alexmogwaiii (approved):
Good job 🔥

src/components/SearchField/SearchField.jsx
```diff
@@ -0,0 +1,55 @@
...
+  state = {
+    query: '',
+  };
+
+  render() {
+    const { movies } = this.props;
+
+    const searchQuery = this.state.query.toLocaleLowerCase().trim();
+    const filteredMovies = [...movies]
+      .filter(movie => movie.title.toLocaleLowerCase().includes(searchQuery)
```
varseniuk, line 15:
```suggestion
    const filteredMovies = movies.filter(movie => 
      movie.title.toLocaleLowerCase().includes(searchQuery)
```
You have no need to create the copy of 'movies' because 'filter' is a non-mutating method

varseniuk (changes requested):
The logic is correct and the solution is good. 
But why did you move the search bar and list into another component? If you wanted to make the task with (*) so you had to make a separate component for the search bar and all its logic and leave 'MoviesList' in App.js (where they initially were).
For this (*) task, you need to know how to lift the state up

varseniuk (approved):
👍

PR #218, opened 2021-01-21
Solution

ShyianO:
[DEMO LINK](https://ShyianO.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,12 +1,34 @@
 import React, { Component } from 'react';
+
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
+    movies: moviesFromServer,
+    query: '',
+  };
+
+  getVisibleMovies = ({ movies, query }) => {
+    let visibleMovies = [...movies];
+
+    if (query) {
+      const lowerQuery = query.toLowerCase();
+
+      visibleMovies = visibleMovies
+        .filter(movie => movie.title.toLowerCase().includes(lowerQuery));
+    }
+
+    return visibleMovies;
+  }
 
   render() {
+    const { movies, query } = this.state;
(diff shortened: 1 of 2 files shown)


khrulenko (approved):

src/App.js
```diff
@@ -1,12 +1,34 @@
...
+  };
+
+  getVisibleMovies = ({ movies, query }) => {
+    let visibleMovies = [...movies];
+
+    if (query) {
+      const lowerQuery = query.toLowerCase();
+
+      visibleMovies = visibleMovies
+        .filter(movie => movie.title.toLowerCase().includes(lowerQuery));
```
varseniuk, line 20: You should also search by description, not only title

> Filter the movies using state.query by searching it in the movie.title or movie.description

src/App.js
```diff
@@ -22,12 +44,18 @@ export class App extends Component {
                   id="search-query"
                   className="input"
                   placeholder="Type search word"
+                  value={query}
+                  onChange={(event) => {
+                    this.setState({
+                      query: event.target.value,
+                    });
```
varseniuk, line 51: Let's make a function (method) for changing state and just call it here

varseniuk (changes requested):

src/App.js
```diff
@@ -1,12 +1,43 @@
...
+  }
+
+  getVisibleMovies = ({ movies, query }) => {
+    let visibleMovies = [...movies];
+
+    if (query) {
+      const lowerQuery = query.toLowerCase();
+
+      visibleMovies = visibleMovies
+        .filter(movie => (
```
varseniuk, line 26:
```suggestion
    if (query) {
      const lowerQuery = query.toLowerCase();

      visibleMovies = movies.filter(movie => (
```
'filter' is a not-mutating method. Why do you create a copy?

varseniuk (approved):
Pay attention to my comment about copying an array

PR #221, opened 2021-01-21
filtered the movies by searching

WhiteRhin0:
- [DEMO LINK](https://WhiteRhin0.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,7 +4,14 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+  };
+
+  filterFunc = movie => (
+    movie.title.toLowerCase().includes(this.state.query.toLowerCase())
+    || movie.description.toLowerCase().includes(this.state.query.toLowerCase())
+  );
 
   render() {
     return (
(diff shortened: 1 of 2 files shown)


5Mountains (approved):
Nice work! 👍

src/App.js
```diff
@@ -22,12 +29,18 @@ export class App extends Component {
...
+                    this.setState({ query: event.target.value });
+                  }}
                 />
               </div>
             </div>
           </div>
 
-          <MoviesList movies={moviesFromServer} />
+          <MoviesList
+            movies={moviesFromServer.filter(this.filterFunc)}
```
varseniuk, line 42: Let's move the expression `moviesFromServer.filter(this.filterFunc)` to separate variable and just pass that variable here

src/App.js
```diff
@@ -22,12 +29,18 @@ export class App extends Component {
                   id="search-query"
                   className="input"
                   placeholder="Type search word"
+                  value={this.state.query}
+                  onChange={(event) => {
+                    this.setState({ query: event.target.value });
```
varseniuk, line 34: Let's create a method like `handleChange` and change the state in it. Here 'onChange' we will just call it

varseniuk (changes requested):

src/App.js
WhiteRhin0: ++

src/App.js
WhiteRhin0: done

varseniuk (approved):
Good job 👍

PR #223, opened 2021-01-21
 react_movies-list-filter 

Nvalja:
[DEMO LINK](https://Nvalja.github.io/react_movies-list-filter/)


xipholena (approved):
Well done!

varseniuk (approved):
Code is good but pls don't spend extra time on changing styles for tasks like this

PR #225, opened 2021-01-21
Solution

Dima-Semenov:
[DEMO LINK](https://Dima-Semenov.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,7 +4,26 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+  };
+
+  searchMovie = () => {
+    const { query } = this.state;
+
+    const filtredMovies = [...moviesFromServer].filter(movie => (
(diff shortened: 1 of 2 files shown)


taniazt (approved):
Nice!

src/App.js
```diff
@@ -4,7 +4,26 @@ import { MoviesList } from './components/MoviesList';
...
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+  };
+
+  searchMovie = () => {
+    const { query } = this.state;
+
+    const filtredMovies = [...moviesFromServer].filter(movie => (
```
varseniuk, line 14:
```suggestion
    const filtredMovies = moviesFromServer.filter(movie => (
```
'filter' is a not-mutating method. Why do you create a copy?

varseniuk (changes requested):

src/App.js
Dima-Semenov: Yes, you are right. My mistake, sorry

varseniuk (approved):
👍

PR #231, opened 2021-01-22
movies Filter

kerpichonka:
[DEMO LINK](https://kerpichonka.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,33 +1,32 @@
 import React, { Component } from 'react';
 import './App.scss';
+import { SearchMovie } from './components/SearchMovie/SearchMovie';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+  };
+
+  filterMovies = (event) => {
+    this.setState({ query: event.target.value });
+  };
 
   render() {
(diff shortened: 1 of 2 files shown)


Serhii-Alieksieievych (approved):

src/App.js
```diff
@@ -1,33 +1,32 @@
...
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+  };
+
+  filterMovies = (event) => {
+    this.setState({ query: event.target.value });
+  };
```
aliceblunt8, line 14: let's pass here event.target.value as an argument

```suggestion
  saveQuery = (value) => {
    this.setState({ query: value });
  };
```

aliceblunt8: also, let's save lowercased value

src/App.js
```diff
@@ -1,33 +1,32 @@
...
 
   render() {
+    const { query } = this.state;
+    const searchMovie = moviesFromServer.filter(movie => movie.title
```
aliceblunt8, line 18:
```suggestion
    const filteredMovies = moviesFromServer.filter(movie => movie.title
```

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #236, opened 2021-01-25
solution

NazarDovhanchuk:
[DEMO LINK](https://nazardovhanchuk.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,9 +4,17 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+  };
 
   render() {
+    const { query } = this.state;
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -4,9 +4,17 @@ import { MoviesList } from './components/MoviesList';
...
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+  };
 
   render() {
+    const { query } = this.state;
+    const filtered = [...moviesFromServer].filter(movie => (
```
aliceblunt8, line 13:
```suggestion
    const filteredMovies = [...moviesFromServer].filter(movie => (
```

aliceblunt8: what does filter method return? (do you really need to make an array copy?)

aliceblunt8 (approved):
@NazarDovhanchuk

PR #308, opened 2021-05-22
I used debounce fn only for training

Ihor-Karpyn:
[DEMO LINK](https://Ihor-Karpyn.github.io/react_movies-list-filter/)


Burunduk003 (approved):
Good job. The unnecessarily complicated solution, but in terms of education, why not?

mykola-yarchuk (approved):
great

PR #314, opened 2021-05-26
Solution v0.1

SkyPKWolf:
[DEMO LINK](https://SkyPKWolf.github.io/react_movies-list-filter/)

### DIFF
--- a/src/components/Search/Search.jsx
+++ b/src/components/Search/Search.jsx
@@ -1,0 +1,53 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { MoviesList } from '../MoviesList';
+
+export class Search extends Component {
+  state = {
+    query: '',
+  };
+
+  render() {
+    const { movies } = this.props;
+    const textToLower = this.state.query.toLocaleLowerCase();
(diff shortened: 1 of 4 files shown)


src/components/Search/Search.jsx
```diff
@@ -0,0 +1,53 @@
...
+
+export class Search extends Component {
+  state = {
+    query: '',
+  };
+
+  render() {
+    const { movies } = this.props;
+    const textToLower = this.state.query.toLocaleLowerCase();
+    const visibleMovies = [...movies].filter(({ title, description }) => (
```
victor-arutiunov, line 13: Filter method returns new array, so you don't need to create a copy of movies.

victor-arutiunov (commented):
Nice work!

src/components/Search/Search.jsx
SkyPKWolf: you are right

varseniuk (approved):
Your logic is correct, but pay attention to the composition of the components. It's good to split the logic and make separate components do things that they are supposed to do (read my comment)

PR #331, opened 2021-05-26
created movie-list-filter

YuriiLiso:
[DEMO LINK](https://yuriiliso.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,7 +14,10 @@ export class App extends Component {
     const visibleMovies = moviesFromServer
       .filter(movie => movie.title
         .toLowerCase()
-        .includes(query.toLowerCase()));
+        .includes(query.toLowerCase())
+        || movie.description
+          .toLowerCase()
+          .includes(query.toLowerCase()));
 
     this.setState({
       query,


rauzok (changes requested):
It does not work correct, I can not write any big letter "A,B,C,D...."

src/App.js
```diff
@@ -4,9 +4,30 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+    visibleMovies: moviesFromServer,
```
varseniuk, line 9:
```suggestion
```
this value is redundant. You can use directly 'moviesFromServer'

src/App.js
```diff
@@ -4,9 +4,30 @@ import { MoviesList } from './components/MoviesList';
...
+        .toLowerCase()
+        .includes(query.toLowerCase())
+        || movie.description
+          .toLowerCase()
+          .includes(query.toLowerCase()));
+
+    this.setState({
+      query,
+      visibleMovies,
+    });
```
varseniuk, line 25:
```suggestion
```
You can get rid of this value in the state and make filtering in `render` method. Then you should only pass `visibleMovies` to the `MoviesList` component

varseniuk (approved):
Good, but pay attention to my comments

PR #346, opened 2021-05-26
test

C-y-y-y:
[DEMO LINK](https://C-y-y-y.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,10 +1,28 @@
 import React, { Component } from 'react';
-import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+import './App.scss';
+
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+    isVisible: moviesFromServer,
+  }
+
+  changeQuery = (event) => {
+    this.setState({
+      query: event.target.value,
+      isVisible: moviesFromServer.filter(
+        movie => movie.title
+          .toLowerCase()
+          .includes(event.target.value.toLowerCase())
+          || movie.description
+            .toLowerCase()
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -1,10 +1,28 @@
...
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+import './App.scss';
+
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+    isVisible: moviesFromServer,
```
YuriiLiso, line 10:
```suggestion
    visibleMovies: moviesFromServer,
```
It will better name for value array.

src/App.js
```diff
@@ -1,10 +1,28 @@
...
+  }
+
+  changeQuery = (event) => {
+    this.setState({
+      query: event.target.value,
+      isVisible: moviesFromServer.filter(
```
YuriiLiso, line 16:
```suggestion
      visibleMovies: moviesFromServer.filter(
```

YuriiLiso (approved):

src/App.js
varseniuk: Actually, there's no need for this value at all.
You can use directly `moviesFromServer`

src/App.js
varseniuk: Good point

varseniuk: As a variant you also can filter the movies inside `render` method (not in the state)

varseniuk (approved):
Good! Pay attention to comments

varseniuk (changes requested):
👍

varseniuk (approved):

PR #347, opened 2021-05-26
first solution

ivan-shpynda:
[DEMO LINK](https://ivan-shpynda.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,12 +1,22 @@
-import React, { Component } from 'react';
+import React from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-export class App extends Component {
-  state = {};
+export class App extends React.Component {
+  state = {
+    query: '',
+  };
+
+  handleQuery = (event) => {
+    this.setState({ query: event.target.value });
+  }
 
   render() {
+    const selectedMovies = moviesFromServer
+      .filter(movie => movie.title.toLowerCase()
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -1,12 +1,22 @@
...
+
+  handleQuery = (event) => {
+    this.setState({ query: event.target.value });
+  }
 
   render() {
+    const selectedMovies = moviesFromServer
+      .filter(movie => movie.title.toLowerCase()
+        .startsWith(this.state.query.toLowerCase()));
+
```
YuriiLiso, line 19:
```suggestion
    const selectedMovies = moviesFromServer
      .filter(movie => movie.title.toLowerCase()
        .startsWith(this.state.query.toLowerCase()));

```
Your code filter only by ```title```. Reread instruction.

YuriiLiso (changes requested):
Good job, but need some changes.

src/App.js
ivan-shpynda: well, there is some ambiguity in the instruction, but i think you are right

src/App.js
```diff
@@ -1,12 +1,22 @@
...
+    query: '',
+  };
+
+  handleQuery = (event) => {
+    this.setState({ query: event.target.value });
+  }
 
   render() {
+    const selectedMovies = moviesFromServer
+      .filter(movie => ((`${movie.title} ${movie.description}`).toLowerCase()
```
varseniuk, line 17: clever 👍

varseniuk (approved):
Good job!

PR #357, opened 2021-05-27
solution

MinchukO:
[DEMO LINK](https://minchuko.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,35 +4,54 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+    findMovies: moviesFromServer,
+  };
 
-  render() {
-    return (
-      <div className="page">
-        <div className="page-content">
-          <div className="box">
-            <div className="field">
-              <label htmlFor="search-query" className="label">
-                Search movie
-              </label>
+changeQuery = (event) => {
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -4,35 +4,54 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export class App extends Component {
-  state = {};
+  state = {
+    query: '',
+    findMovies: moviesFromServer,
```
varseniuk, line 9:
```suggestion
```
Actually, this value is redundant. You can use directly `moviesFromServer`

src/App.js
```diff
@@ -4,35 +4,54 @@ import { MoviesList } from './components/MoviesList';
...
-        <div className="page-content">
-          <div className="box">
-            <div className="field">
-              <label htmlFor="search-query" className="label">
-                Search movie
-              </label>
+changeQuery = (event) => {
+  this.setState({
+    query: event.target.value,
+    findMovies: moviesFromServer.filter(
```
varseniuk, line 15: You always use the same array - `moviesFromServer` (without any changes) so you may filter it in method `render` directly without writing it to the state. Just create variable `findMovies`  in 'render', filter there `MoviesFromServer` and pass it to the `MoviesList`

varseniuk (approved):
Good! 
Pay attention to my comments

PR #360, opened 2021-05-27
Solution-001

senkiv-oleh:
[DEMO LINK](https://senkiv-oleh.github.io/react_movies-list-filter/)


varseniuk (approved):
Good job with lifting the state up! 👍

PR #425, opened 2021-09-03
Develop

lera-yaskevych:
[DEMO LINK](https://lera-yaskevych.github.io/react_movies-list-filter/)
I couldn't figure out how to move the logic from App to SearchBar component

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
+import { SearchBar } from './components/SearchBar/SearchBar';
 import moviesFromServer from './api/movies.json';
 
 export class App extends React.Component<{}, State> {
@@ -19,30 +20,14 @@ export class App extends React.Component<{}, State> {
     const { movies, query } = this.state;
     const filteredMovies = movies
       .filter(movie => (
-        movie.title.toLowerCase().includes(query.toLowerCase())
-        || movie.description.toLowerCase().includes(query.toLowerCase())
+        (movie.title + movie.description).toLowerCase()
+          .includes(query.toLowerCase())
       ));
 
     return (
       <div className="page">
         <div className="page-content">
-          <div className="box">
-            <div className="field">
-              <label htmlFor="search-query" className="label">
-                Search movie
-              </label>
-
-              <div className="control">
-                <input
-                  type="text"
-                  id="search-query"
-                  className="input"
-                  placeholder="Type search word"
-                  onChange={this.changeHandler}
-                />
-              </div>
-            </div>
-          </div>
+          <SearchBar queryUpdateFunction={this.changeHandler} />
 
           <MoviesList movies={filteredMovies} />
         </div>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,12 +3,26 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-type State = {};
-
 export class App extends React.Component<{}, State> {
-  state: State = {};
+  state: State = {
+    query: '',
+    movies: moviesFromServer,
+  };
+
+  changeHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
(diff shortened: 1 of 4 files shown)


src/react-app-env.d.ts
```diff
@@ -1,5 +1,10 @@
 /// <reference types="react-scripts" />
 
+type State = {
+  query: string;
+  movies: Movie[];
+};
+
```
illiarerun, line 7: You should move it to file with component in which this state is used

src/App.tsx
```diff
@@ -1,35 +1,35 @@
...
+    this.setState({
+      query: event.target.value,
+    });
+  };
 
   render() {
+    const { movies, query } = this.state;
+    const filteredMovies = movies
+      .filter(movie => (
+        (movie.title + movie.description).toLowerCase()
```
illiarerun, line 23: Not really sure if concatenation of this string is a good idea, but it's subjective)

illiarerun (approved):
Great job 🔥 Got only two little comments about code

src/App.tsx
lera-yaskevych: Not the best solution, I do agree on that. It just made this line of code much shorter😅 
But if I understand the concepts right, the time complexity would be equal in both cases, since string concatenation takes constant time. But I should check on that on Q&A. Google keeps giving me different answers

lera-yaskevych:
> Great job 🔥 Got only two little comments about code

Thanks! You gave me some points to consider)

src/App.tsx
```diff
@@ -1,35 +1,46 @@
...
+  state: State = {
+    query: '',
+    movies: moviesFromServer,
+  };
+
+  getFilteredMovies = () => {
+    const { movies, query } = this.state;
+    const filteredMovies = movies
+      .filter(movie => (
+        (movie.title + movie.description).toLowerCase()
```
Denchepornyuk, line 22: Screenshot 2021-09-10 at 12 42 03

з таким записом буде проходити ось такий search, а він не мав пройти

Denchepornyuk (changes requested):

src/App.tsx
lera-yaskevych: Протуп. Дійсно, зараз справлю 👌

src/App.tsx
```diff
@@ -1,35 +1,46 @@
...
+    query: '',
+    movies: moviesFromServer,
+  };
+
+  getFilteredMovies = () => {
+    const { movies, query } = this.state;
+    const filteredMovies = movies
+      .filter(movie => (
+        (movie.title.toLowerCase().includes(query.toLowerCase())
+        || movie.description.toLowerCase().includes(query.toLowerCase())
```
danheim, line 23: you do `query.toLowerCase` twice, do once

danheim (changes requested):

src/App.tsx
```diff
@@ -1,35 +1,47 @@
...
+  };
+
+  getFilteredMovies = () => {
+    const { movies, query } = this.state;
+    const lowercaseQuery = query.toLowerCase();
+    const filteredMovies = movies
+      .filter(movie => (
+        (movie.title.toLowerCase().includes(lowercaseQuery)
+        || movie.description.toLowerCase().includes(lowercaseQuery)
+        )));
```
danheim, line 25: more human readable
```suggestion
const filteredMovies = movies.filter(movie => {
  const { title, description } = movie;

  const conditions = [
    title.toLowerCase().includes(lowercaseQuery),
    description.toLowerCase().includes(lowercaseQuery),
  ];

  return conditions.some(Boolean);
});
```

src/components/SearchBar/SearchBar.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+
+type Props = {
+  queryUpdateFunction: (event: React.ChangeEvent<HTMLInputElement>) => void;
```
danheim, line 4:
```suggestion
  onChangeCallback: (event: React.ChangeEvent<HTMLInputElement>) => void;
```

danheim (approved):

PR #427, opened 2021-09-04
Develop

illiarerun:
-  [DEMO LINK](https://illiarerun.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -13,13 +13,3 @@
   padding: 20px;
   border-left: 1px solid lightgray;
 }
-
-.movies {
-  display: grid;
-  grid-template-columns: repeat(auto-fill, minmax(250px, 1fr));
-  gap: 20px;
-}
-
-.card-image img {
-  object-fit: contain;
-}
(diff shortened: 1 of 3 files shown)


alex-glotov67 (approved):
Not sure that movieList need move to your component `SearchBox`, because in Task we have a point about it. But it still looks good!

illiarerun:
> Not sure that movieList need move to your component `SearchBox`, because in Task we have a point about it. But it still looks good!

Yeah, that's why i've commented about moving logic to SearchBox. If we need to move search logic to another component than we are supposed to move filter function, right? But how can we do it without passing movies to filter to it as a props?))

alex-glotov67:
> > Not sure that movieList need move to your component `SearchBox`, because in Task we have a point about it. But it still looks good!
> 
> Thanks for your review 🔥
> Yeah, that's why i've commented about moving logic to SearchBox. If we need to move search logic to another component than we are supposed to move filter function, right? But how can we do it without passing movies to filter to it as a props?))

Its describes in the next topic =)

Denchepornyuk (approved):

PR #428, opened 2021-09-06
Solution

alex-glotov67:
- [DEMO LINK](https://alex-glotov67.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,34 +2,34 @@ import React from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+import { SearchField } from './components/SearchField';
 
-type State = {};
+type State = {
+  query: string;
+};
 
 export class App extends React.Component<{}, State> {
-  state: State = {};
+  state: State = {
+    query: '',
+  };
+
+  setQuery = (query: string) => {
+    this.setState({ query });
+  };
 
   render() {
+    const { query } = this.state;
+
+    const newFilms = [...moviesFromServer].filter(({ title, description }) => {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -2,34 +2,34 @@ import React from 'react';
...
+
+  setQuery = (query: string) => {
+    this.setState({ query });
+  };
 
   render() {
+    const { query } = this.state;
+
+    const filteredMovies = [...moviesFromServer].filter(({ title, description }) => {
+      return description.includes(query) || title.toUpperCase().includes(query.toUpperCase());
```
lera-yaskevych, line 24:
```suggestion
      return description.toUpperCase().includes(query.toUpperCase()) || title.toUpperCase().includes(query.toUpperCase());
```
I guess it would work better like this, so whenever some couch potato like me, who doesn't use uppercase that much often to google things, wants to find a movie which takes place in New York it would be easier for them

src/App.tsx
```diff
@@ -2,34 +2,34 @@ import React from 'react';
...
 
 export class App extends React.Component<{}, State> {
-  state: State = {};
+  state: State = {
+    query: '',
+  };
+
+  setQuery = (query: string) => {
+    this.setState({ query });
+  };
```
lera-yaskevych, line 18: Iteresting approach! I like it

lera-yaskevych (approved):
Good job! I have just one comment

src/App.tsx
alex-glotov67: Good point! Thanks! 🚀

ArtemTopolskyi (approved):

PR #436, opened 2021-09-08
React Movies list filter solution by Holiuk Taras

TarasHoliuk:
[DEMO LINK](https://tarasholiuk.github.io/react_movies-list-filter/)

Попереднє рішення (initial commit) без використання async (з двома аргументами в setState) який спосіб кращий?

### DIFF
--- a/src/components/Search/Search.tsx
+++ b/src/components/Search/Search.tsx
@@ -19,12 +19,14 @@ export class Search extends React.Component<Props, State> {
   handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
     const { value } = event.target;
 
-    this.setState(() => ({
+    this.setState({
       query: value,
-    }), this.getVisibleMovies);
+    });
   };
 
-  getVisibleMovies = () => {
+  getVisibleMovies = async (event: React.ChangeEvent<HTMLInputElement>) => {
+    await this.handleChange(event);
+
     const { movies } = this.props;
     const { query } = this.state;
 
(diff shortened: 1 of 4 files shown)


SashaLifashkin:
Хорошая работа!

src/components/MoviesList/MoviesList.tsx
```diff
@@ -4,12 +4,19 @@ import { MovieCard } from '../MovieCard';
...
-  <div className="movies">
-    {movies.map(movie => (
-      <MovieCard key={movie.imdbId} movie={movie} />
-    ))}
-  </div>
-);
+export const MoviesList: React.FC<Props> = ({ movies, query }) => {
+  const visibleMovies = movies
+    .filter(movie => movie.title.toLowerCase().includes(query.toLowerCase())
+        || movie.description.includes(query.toLowerCase()));
```
Elena-brv, line 13: query.toLowerCase() better to pass in a separate variable not to calculate it in every iteration

src/components/Search/Search.tsx
```diff
@@ -0,0 +1,54 @@
...
+    const { value } = event.target;
+
+    this.setState({
+      query: value,
+    });
+  };
+
+  render() {
+    const { query } = this.state;
+    const { movies } = this.props;
```
Elena-brv, line 27: you don't use movies in this component, you just pass it on. So it's better to render both components (MoviesList and Search) in the App

Elena-brv (changes requested):

src/components/Search/Search.tsx
TarasHoliuk: yes, I should use lifting state up or render components in the App

Elena-brv (approved):

PR #442, opened 2021-09-10
Solution

Crystallord:
[DEMO LINK](https://crystallord.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,12 +3,31 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-type State = {};
+type State = {
+  moviesList: Movie[],
+  query: string;
+};
 
 export class App extends React.Component<{}, State> {
-  state: State = {};
+  state: State = {
+    moviesList: moviesFromServer,
+    query: '',
+  };
+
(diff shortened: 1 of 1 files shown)


DmytroMykoliv (approved):
Good work, easy to understand ) 👍

src/App.tsx
```diff
@@ -3,12 +3,31 @@ import './App.scss';
...
+};
 
 export class App extends React.Component<{}, State> {
-  state: State = {};
+  state: State = {
+    moviesList: moviesFromServer,
+    query: '',
+  };
+
+  filteredList = (movie: Movie) => {
```
Elena-brv, line 17: not meaningful name of the method. Better to call it "shouldBeFiltered" or something like that.

src/App.tsx
```diff
@@ -3,12 +3,31 @@ import './App.scss';
...
-type State = {};
+type State = {
+  moviesList: Movie[],
+  query: string;
+};
 
 export class App extends React.Component<{}, State> {
-  state: State = {};
+  state: State = {
+    moviesList: moviesFromServer,
```
Elena-brv, line 13: you don't need this property on state

Elena-brv (approved):

PR #501, opened 2021-10-19
solution

Saharij:
[DEMO](https://Saharij.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -28,7 +28,7 @@ export class App extends React.Component<{}, State> {
       <div className="page">
         <div className="page-content">
 
-          <Search onChangeFunction={this.onChangeFunction} />
+          <Search value={this.state.query} onChange={this.onChangeFunction} />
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,35 +1,35 @@
...
+    query: '',
+  };
+
+  onChangeFunction = (value: string): void => {
+    this.setState({ query: value });
+  };
 
   render() {
+    const { query } = this.state;
+    const movies = [...moviesFromServer].filter(({ title, description }) => (
```
aliceblunt8, line 22: what does filter return? is it mutating method?
```suggestion
    const movies = moviesFromServer.filter(({ title, description }) => (
```

aliceblunt8 (approved):
GJ

src/App.tsx
Saharij: you are right

PR #507, opened 2021-10-19
v1

roman-toritsyn:
[DEMO LINK](https://roman-toritsyn.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,18 +5,16 @@ import moviesFromServer from './api/movies.json';
 
 type State = {
   query: string,
-  movies: Movie[],
 };
 
 export class App extends React.Component<{}, State> {
   state: State = {
     query: '',
-    movies: [...moviesFromServer],
   };
 
   render() {
     const { query } = this.state;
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -3,33 +3,51 @@ import './App.scss';
...
+type State = {
+  query: string,
+  movies: Movie[],
+};
 
 export class App extends React.Component<{}, State> {
-  state: State = {};
+  state: State = {
+    query: '',
+    movies: [...moviesFromServer],
```
aliceblunt8, line 14: no need to save movies in state, just use moviesFromServer for filtering
```suggestion
```

src/App.tsx
```diff
@@ -3,33 +3,51 @@ import './App.scss';
...
+  };
 
   render() {
+    const { query } = this.state;
+    const movies = [...this.state.movies].filter(
```
aliceblunt8, line 19: no need to make copy for filter, it doesn't mutate array anyway

```suggestion
    const filteredmMovies = moviesFromServer.filter(
```

PR #510, opened 2021-10-19
solution

DY-JS:
[DEMO LINK](https://DY-JS.github.io/react_movies-list-filter/)


src/App.tsx
```diff
@@ -3,12 +3,35 @@ import './App.scss';
...
+  query: string;
+};
 
 export class App extends React.Component<{}, State> {
-  state: State = {};
+  state: State = {
+    query: '',
+  };
+
+  searchHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
```
aliceblunt8, line 15:
```suggestion
  handleSearch = (event: React.ChangeEvent<HTMLInputElement>) => {
```

src/App.tsx
```diff
@@ -3,12 +3,35 @@ import './App.scss';
...
+    this.setState({
+      query: event.currentTarget.value,
+    });
+  };
+
+  sortHandler = () => {
```
aliceblunt8, line 21: you filter movies, you don't sort them

```suggestion
  getFilteredMovies = () => {
```

src/App.tsx
```diff
@@ -3,12 +3,35 @@ import './App.scss';
...
+    const movies = moviesFromServer;
+    const { query } = this.state;
+
+    return (
+      movies.filter(item => item.title.toLowerCase()
```
aliceblunt8, line 26: no need for movies variables, just use moviesFromServer
```suggestion
    const { query } = this.state;

    return (
      moviesFromServer.filter(item => item.title.toLowerCase()
```

src/App.tsx
```diff
@@ -3,12 +3,35 @@ import './App.scss';
...
+        .includes(query.toLowerCase())
+        || item.description.toLowerCase().includes(query.toLowerCase()))
+    );
+  };
 
   render() {
+    const sortedMovies = this.sortHandler();
```
aliceblunt8, line 33:
```suggestion
    const filteredMovies = this.sortHandler();
```

PR #527, opened 2021-11-11
code added

yatochkaaa:
[DEMO LINK](https://yatochkaaa.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,37 @@
+/* eslint-disable jsx-a11y/label-has-associated-control */
 import React from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-type State = {};
+type State = {
+  query: string;
+};
+
+interface Event {
+  target: {
+    value: string;
+  };
+}
 
 export class App extends React.Component<{}, State> {
-  state: State = {};
+  state: State = {
+    query: '',
+  };
+
+  handleChange = (event: Event) => (
+    this.setState({ query: event.target.value }));
 
   render() {
+    const { query } = this.state;
+
+    const visibleMovies = [...moviesFromServer].filter(
+      movie => (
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,14 +1,37 @@
...
-type State = {};
+type State = {
+  query: string;
+};
+
+interface Event {
+  target: {
+    value: string;
+  };
+}
```
Olexandr6, line 15: You shoud use `React.ChangeEvent<HTMLInputElement>` type instead of this custom interface👆

Olexandr6 (changes requested):
Check comments👇

src/App.tsx
```diff
@@ -1,14 +1,31 @@
...
 
   render() {
+    const { query } = this.state;
+
+    const visibleMovies = moviesFromServer.filter(
+      movie => (
+        movie.title.toLowerCase().includes(query.toLocaleLowerCase())
+        || movie.description.toLowerCase().includes(query.toLocaleLowerCase())
+      ),
+    );
```
Olexandr6, line 27: Take this logic to a separate method and call it in the render method. For example: `getVisibleMovies()`

src/App.tsx
```diff
@@ -20,16 +37,18 @@ export class App extends React.Component<{}, State> {
...
                   id="search-query"
                   className="input"
                   placeholder="Type search word"
                 />
               </div>
             </div>
           </div>
-
-          <MoviesList movies={moviesFromServer} />
+          <MoviesList movies={visibleMovies} />
```
Olexandr6, line 51: And here you can use this method
```suggestion
          <MoviesList movies={this.getVisibleMovies()} />
```

src/App.tsx
VinnieJ-2k20: Agree, good advice.

VinnieJ-2k20 (approved):
Good job 👍 
Check comments for some code optimisation tips.

PR #544, opened 2021-11-15
[REACT] Movies list filter STRZH

DionStrzh:
[DEMO LINK](https://dionstrzh.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,34 @@
+/* eslint-disable jsx-a11y/label-has-associated-control */
 import React from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-type State = {};
+type State = {
+  query: string;
+};
 
 export class App extends React.Component<{}, State> {
-  state: State = {};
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,14 +1,34 @@
...
+    const { query } = this.state;
+    const visibleMovies = moviesFromServer.filter((movie) => {
+      const movieTitle = movie.title.toLowerCase();
+      const movieDescription = movie.description.toLowerCase();
+
+      return (
+        movieTitle.includes(query.toLowerCase())
+        || movieDescription.includes(query.toLowerCase())
+      );
+    });
```
Olexandr6, line 30: Take this logic to a separate method and call it in the render method. For example: getVisibleMovies().

Olexandr6 (approved):
Good solution👍 Check comments👇

src/App.tsx
VinnieJ-2k20: Yes, good advice.

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment on code optimisation.

PR #546, opened 2021-11-15
solution

olena1994:
[DEMO LINK](https://olena1994.github.io/react_movies-list-filter/)


dima-narozhnyi (approved):
Good solution 👍

src/App.tsx
```diff
@@ -1,14 +1,28 @@
...
 
   render() {
+    const { query } = this.state;
+
+    const visibleMovies = moviesFromServer.filter(
+      movie => (
+        movie.title.toLowerCase().includes(query.toLowerCase())
+        || movie.description.toLowerCase().includes(query.toLowerCase())
+      ),
+    );
```
VinnieJ-2k20, line 24: It would be great to move filtering to class method and just call it here.
This way `render` method stays more simple.

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment.

PR #550, opened 2021-11-16
solution

ivarovskiy:
[DEMO LINK](https://ivarovskiy.github.io/react_movies-list-filter/)


Igor-MateAcademy (approved):
All right, Abobus, but i can't to enter letters in uppercase)

VinnieJ-2k20 (changes requested):
User should be able to type uppercase letters, but filter should be not case-sensitive.

VinnieJ-2k20 (approved):
Good job 👍

PR #554, opened 2021-11-26
Implement search bar

dmtrhrytsak:
[List filter](https://dmtrhrytsak.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,56 @@
 import React from 'react';
 import './App.scss';
+
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+import './types/Movie';
 
-type State = {};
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,56 @@
...
+    });
+  };
 
   render() {
+    const { query } = this.state;
+    const queryRegExp = new RegExp(query, 'i');
+
+    const visibleMovies: Movie[] = moviesFromServer.filter(
+      (movie) => queryRegExp.test(movie.title) || queryRegExp.test(movie.description),
+    );
```
aionova, line 29: I don't really see any sense in using regExp, when you could simply use 'includes'. But it's up to you

aionova (approved):
GJ!

PR #559, opened 2021-12-13
Movies list search

Vitaliy-0:
[DEMO](https://vitaliy-0.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,40 @@
+/* eslint-disable jsx-a11y/label-has-associated-control */
 import React from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-type State = {};
+type State = {
+  query: string,
+};
 
 export class App extends React.Component<{}, State> {
-  state: State = {};
+  state: State = {
+    query: '',
+  };
+
+  handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+
+    this.setState({ query: value });
+  };
 
   render() {
+    const { query } = this.state;
+
+    const visibleMovies = moviesFromServer.filter(movie => {
+      const lowerTitle = movie.title.toLowerCase();
+      const lowerDescription = movie.description.toLowerCase();
+      const lowerQuery = query.toLowerCase();
+
+      if (lowerTitle.includes(lowerQuery)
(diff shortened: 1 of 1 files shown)


msaikun (approved):
GJ!

ihor-jpeg (changes requested):
I think you misunderstood the task description. There should be a query in your state and onChange should be working with the query, not the array of movies

src/App.tsx
```diff
@@ -1,14 +1,40 @@
...
+      const lowerTitle = movie.title.toLowerCase();
+      const lowerDescription = movie.description.toLowerCase();
+      const lowerQuery = query.toLowerCase();
+
+      if (lowerTitle.includes(lowerQuery)
+        || lowerDescription.includes(lowerQuery)) {
+        return movie;
+      }
+
+      return undefined;
```
ihor-jpeg, line 35: wait, what? I think there is a better way to write the return value of this callback. do not return undefined explicitly.

ihor-jpeg (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,40 @@
...
+      const lowerTitle = movie.title.toLowerCase();
+      const lowerDescription = movie.description.toLowerCase();
+      const lowerQuery = query.toLowerCase();
+
+      if (lowerTitle.includes(lowerQuery)
+        || lowerDescription.includes(lowerQuery)) {
+        return movie;
+      }
+
+      return null;
```
ihor-jpeg, line 35: filter's callback should return true or false. 
why not just use the result of the includes method? 

```suggestion
      return lowerTitle.includes(lowerQuery) || lowerDescription.includes(lowerQuery);
```

ihor-jpeg (changes requested):
well done! but let's fix your filter method

src/App.tsx
```diff
@@ -1,14 +1,36 @@
...
+    const { query } = this.state;
+
+    const visibleMovies = moviesFromServer.filter(movie => {
+      const lowerTitle = movie.title.toLowerCase();
+      const lowerDescription = movie.description.toLowerCase();
+      const lowerQuery = query.toLowerCase();
+
+      return lowerTitle.includes(lowerQuery)
+      || lowerDescription.includes(lowerQuery);
+    });
```
AlinaMatuschak, line 32: better create method for this logic

AlinaMatuschak (approved):

PR #560, opened 2021-12-13
Solution

msaikun:
- [DEMO LINK](https://msaikun.github.io/react_movies-list-filter/)


src/App.tsx
```diff
@@ -3,33 +3,61 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-type State = {};
+interface Movie {
+  title: string,
+  description: string,
+  imgUrl: string,
+  imdbUrl: string,
+  imdbId: string,
```
Vitaliy-0, line 11: you dont need these fields

src/App.tsx
```diff
@@ -3,33 +3,61 @@ import './App.scss';
...
+    moviesList: moviesFromServer,
+    searchParam: '',
+  };
 
   render() {
+    const { moviesList, searchParam } = this.state;
+
+    let visibleMovies = moviesList;
+
+    if (searchParam.length > 0) {
```
Vitaliy-0, line 30: you really need this check?
You can use like:
const visibleMovies = moviesList.filter...
If length of searchParam is 0 then you will see all of possible movies.

Vitaliy-0 (approved):
Good job!

src/App.tsx
ihor-jpeg: agreed

ihor-jpeg (approved):
Well done. let's just remove that unnecessary if statement

PR #567, opened 2021-12-14
Solution

volodymyr-soltys97:
[DEMO LINK](https://volodymyr-soltys97.github.io/react_movies-list-filter/)


ihor-jpeg (approved):
well done

PR #568, opened 2021-12-14
dev

skachkov23:
[DEMO](https://skachkov23.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,29 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-type State = {};
+type State = {
+  query: string,
+};
 
-export class App extends React.Component<{}, State> {
-  state: State = {};
+export class App extends Component<{}, State> {
+  state: State = {
+    query: '',
+  };
+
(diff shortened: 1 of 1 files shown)


volodymyr-soltys97 (approved):
LGTM 🔥

src/App.tsx
```diff
@@ -1,35 +1,51 @@
...
+  onChangeHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
+    this.setState({ query: event.target.value });
+  };
 
   render() {
+    const { query } = this.state;
+
+    const filteredMovies = moviesFromServer.filter(
+      movie => movie.title.toLowerCase()
+        .includes(query.toLowerCase()),
```
ihor-jpeg, line 24: you are only checking whether the title contains the query, but you need to also check the description

ihor-jpeg (changes requested):
good work, but let's also check the description for the query

src/App.tsx
skachkov23: Sorry, I thought that's optional in this task ("or" in README and in description of task)

ihor-jpeg: Yeah, it was optional. My bad

ihor-jpeg (approved):
Well done!

PR #570, opened 2021-12-14
implemented task

tmrsl:
[DEMO LINK](https://tmrsl.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,12 +3,37 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-type State = {};
+type State = {
+  query: string,
+};
 
 export class App extends React.Component<{}, State> {
-  state: State = {};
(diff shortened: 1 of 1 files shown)


aldmpr (approved):
good!

src/App.tsx
```diff
@@ -3,33 +3,58 @@ import './App.scss';
...
+};
 
 export class App extends React.Component<{}, State> {
-  state: State = {};
+  state: State = {
+    query: '',
+  };
+
+  handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
```
ihor-jpeg, line 16: 👍

ihor-jpeg (changes requested):
Good work! But if my query is in uppercase then the search doesn't work
Screenshot 2021-12-15 at 12 29 00

ihor-jpeg (approved):

PR #605, opened 2022-01-23
solution

samson-sim:
[DEMO LINK](https://samson-sim.github.io/react_movies-list-filter/)


marina-tilniak (approved):

src/App.tsx
```diff
@@ -3,33 +3,50 @@ import './App.scss';
...
-  state: State = {};
+  state: State = {
+    query: '',
+  };
+
+  filterCallback = (movie: Movie) => {
+    return (
+      movie.title.toLowerCase().includes(this.state.query.toLowerCase())
+      || movie.description.toLowerCase().includes(this.state.query.toLowerCase())
+    );
```
Olexandr6, line 19: You can optimize this
```suggestion
  filterCallback = (movie: Movie) => {
  	const { title, description } = movie;
  	const lowerCaseQuery = query.toLowerCase();
  	
    return (
      title.toLowerCase().includes(lowerCaseQuery)
      || description.toLowerCase().includes(lowerCaseQuery)
    );
```

src/App.tsx
```diff
@@ -3,33 +3,50 @@ import './App.scss';
...
+                  Search Movie
+                  <input
+                    type="text"
+                    id="search-query"
+                    value={this.state.query}
+                    onChange={(event) => {
+                      this.setState({
+                        query: event.target.value,
+                      });
+                    }}
```
Olexandr6, line 41: Create a separate handler function for this logic

Olexandr6 (changes requested):
Good solution 🔥 But some things should be optimized

Olexandr6 (approved):
🔥

PR #611, opened 2022-01-25
My solution

SavelovSerhii:
https://savelovserhii.github.io/react_movies-list-filter/


src/App.tsx
```diff
@@ -3,33 +3,51 @@ import './App.scss';
...
 
   render() {
+    const visibleMovies = moviesFromServer.filter((movie) => {
+      if (movie.title.toLowerCase().includes(this.state.query.toLowerCase())
+        || movie.description.toLowerCase().includes(this.state.query.toLowerCase())) {
+        return true;
+      }
+
+      return false;
+    });
```
Basil-Panasiuk, line 23: краще винести в окрему функцію

Basil-Panasiuk (approved):

src/App.tsx
Olexandr6: Agree

src/App.tsx
```diff
@@ -3,33 +3,51 @@ import './App.scss';
...
+                  <input
+                    type="text"
+                    id="search-query"
+                    className="input"
+                    placeholder="Type search word"
+                    value={this.state.query}
+                    onChange={(event) => {
+                      this.setState({
+                        query: event.target.value,
+                      });
```
Olexandr6, line 42: Create a separate handler function for this

src/App.tsx
Olexandr6, line 23: You can simplify this
```suggestion
    const visibleMovies = [...moviesFromServer].filter(({ title, description })  => {
      const lowerCaseQuery = query.toLowerCase();
      
      return title.toLowerCase().includes(lowerCaseQuery)
        || description.toLowerCase().includes(lowerCaseQuery);
    });
```

Olexandr6: Create a separate function for this logic, as well.

Olexandr6 (changes requested):
Check comments 👇

Olexandr6 (approved):

PR #626, opened 2022-01-26
sm msg

Mikedacheka:
Thank you for review!
Have one general question:
movies={this.filterMovies()} - Can we do like that?


AlexAzalor (approved):

src/App.tsx
```diff
@@ -3,33 +3,52 @@ import './App.scss';
...
+    searchWord: '',
+  };
+
+  filteredMovies = () => {
+    const { searchWord } = this.state;
+
+    return moviesFromServer.filter(
+      movie => movie.title.toLowerCase().includes(searchWord.toLocaleLowerCase())
+      || movie.description.toLowerCase().includes(searchWord.toLocaleLowerCase()),
+    );
```
Olexandr6, line 21: You can optimize this
```suggestion
    const lowerCaseSearchWord = this.state.searchWord.toLocaleLowerCase();

    return moviesFromServer.filter(
      movie => movie.title.toLowerCase().includes(lowerCaseSearchWord)
      || movie.description.toLowerCase().includes(lowerCaseSearchWord),
    );
```

src/App.tsx
```diff
@@ -3,33 +3,52 @@ import './App.scss';
...
               <label htmlFor="search-query" className="label">
                 Search movie
+                <div className="control">
+                  <input
+                    type="text"
+                    id="search-query"
+                    onChange={(event) => {
+                      this.setState({
+                        searchWord: event?.target.value,
+                      });
```
Olexandr6, line 41: Create a separate handler function for this logic

Olexandr6 (changes requested):
Some small things should be optimized

Denchepornyuk (approved):

PR #628, opened 2022-01-26
Task is done

ArtemKa2208:
[DEMO LINK](https://ArtemKa2208.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,12 +3,35 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-type State = {};
+type State = {
+  query: string,
+};
 
 export class App extends React.Component<{}, State> {
-  state: State = {};
+  state: State = {
+    query: '',
+  };
+
+  handlerInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    this.setState({
+      query: event.target.value,
+    });
+  };
 
   render() {
+    const { query } = this.state;
+    const visibleMovies = moviesFromServer.filter((value) => {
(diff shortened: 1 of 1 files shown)


JHesher (approved):

src/App.tsx
```diff
@@ -3,33 +3,57 @@ import './App.scss';
...
+      const title = value.title.toLowerCase();
+      const description = value.description.toLowerCase();
+      const lowerCaseQuery = query.toLowerCase();
+
+      if (title.includes(lowerCaseQuery) || description.includes(lowerCaseQuery)) {
+        return true;
+      }
+
+      return false;
+    });
```
danheim, line 33: create getVisibleMovies method
```suggestion
    const visibleMovies = this.getVisibleMovies(moviesFromServer);
```

danheim (changes requested):

LICENSE
```diff
@@ -0,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
```
Elena-brv, line 1: remove this file

src/App.tsx
```diff
@@ -3,33 +3,61 @@ import './App.scss';
...
+  };
+
+  getVisibleMovies = (movies: Movie[]) => {
+    const { query } = this.state;
+
+    return movies.filter((value) => {
+      const title = value.title.toLowerCase();
+      const description = value.description.toLowerCase();
+      const lowerCaseQuery = query.toLowerCase();
```
Elena-brv, line 27: no need to do query to lowercase on every iteration, you can do it before filtering, because query doesn't change

Elena-brv (changes requested):

PR #633, opened 2022-01-26
solution films

OleksandrOstrovskyi:
[DEMO LINK](https://OleksandrOstrovskyi.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -18,16 +18,20 @@ export class App extends React.Component<{}, State> {
     });
   };
 
-  render() {
-    // eslint-disable-next-line no-console
-    console.log(this.state.query);
+  getVisibleMovies = () => {
+    const lowerCaseQuery = this.state.query.toLowerCase();
 
+    return (
+      moviesFromServer.filter(({ title, description }) => {
+        return title.toLowerCase().includes(lowerCaseQuery)
+        || description.toLowerCase().includes(lowerCaseQuery);
+      }));
+  };
+
+  render() {
     const { query } = this.state;
 
-    const visibleMovies = moviesFromServer.filter(movie => (
-      movie.title.toLowerCase().includes(query.toLowerCase())
-      || movie.description.toLowerCase().includes(query.toLowerCase())
-    ));
+    const visibleMovies = this.getVisibleMovies();
 
     return (
       <div className="page">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -3,33 +3,51 @@ import './App.scss';
...
+
+  changeHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
+    this.setState({
+      query: event.target.value,
+    });
+  };
 
   render() {
+    // eslint-disable-next-line no-console
+    console.log(this.state.query);
```
gnosis-frog, line 23: don't leave console.log in prod

gnosis-frog (approved):

src/App.tsx
Olexandr6: Agree, delete it

src/App.tsx
```diff
@@ -3,33 +3,51 @@ import './App.scss';
...
+
+    const { query } = this.state;
+
+    const visibleMovies = moviesFromServer.filter(movie => (
+      movie.title.toLowerCase().includes(query.toLowerCase())
+      || movie.description.toLowerCase().includes(query.toLowerCase())
+    ));
```
Olexandr6, line 30: Simplify this
```suggestion
    const visibleMovies = moviesFromServer.filter({ title, description } => {
      	const lowerCaseQuery = this.state.query.toLowerCase();
      	
        return title.toLowerCase().includes(lowerCaseQuery)
          || description.toLowerCase().includes(lowerCaseQuery)
      });
```

Olexandr6: Create a separate function `getVisibleMovies()` for this logic

Olexandr6 (changes requested):
Check comments 👇

src/App.tsx
```diff
@@ -3,33 +3,55 @@ import './App.scss';
...
+  };
+
+  getVisibleMovies = () => {
+    const lowerCaseQuery = this.state.query.toLowerCase();
+
+    return (
+      moviesFromServer.filter(({ title, description }) => {
+        return title.toLowerCase().includes(lowerCaseQuery)
+        || description.toLowerCase().includes(lowerCaseQuery);
+      }));
```
Olexandr6, line 28: 🔥

Olexandr6 (approved):
Well done 👍

PR #642, opened 2022-01-27
react_movies-list-filter

ViktorBachynskyi:


Sesile26 (approved):

src/App.tsx
```diff
@@ -3,33 +3,46 @@ import './App.scss';
...
+  state: State = {
+    query: '',
+  };
 
   render() {
+    const { query } = this.state;
+    const visibleMovies = moviesFromServer.filter(
+      movie => movie.title.toLowerCase().includes(query.toLowerCase())
+    || movie.description.toLowerCase().includes(query.toLowerCase()),
+    );
```
Elena-brv, line 20: do not write logic in render, create a method;
also create for `query.toLowerCase` a variable on that method, no need to calculate it on every iteration

src/App.tsx
```diff
@@ -3,33 +3,46 @@ import './App.scss';
...
+                <div className="control">
+                  <input
+                    type="text"
+                    id="search-query"
+                    className="input"
+                    placeholder="Type search word"
+                    value={query}
+                    onChange={(event) => {
+                      this.setState({ query: event.target.value });
+                    }}
```
Elena-brv, line 38: also create a method for it

Elena-brv (changes requested):

src/App.tsx
```diff
@@ -3,33 +3,56 @@ import './App.scss';
...
+  };
+
+  inputHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
+    this.setState({
+      query: event.target.value,
+    });
+  };
+
+  filterMovies = (movies: Movie[]) => {
+    this.state.query.toLowerCase();
```
Elena-brv, line 22: you need to set it to some variable, now it not changes the query
```suggestion
    const lowerQuery = this.state.query.toLowerCase();
```

Elena-brv (changes requested):

Denchepornyuk (approved):

PR #643, opened 2022-01-29
solution

Rom911:
[Demo](https://rom911.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,10 +3,30 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-type State = {};
+type State = {
+  query: string,
+};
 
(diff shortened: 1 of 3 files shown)


Rozdorozhnii (approved):

src/App.tsx
```diff
@@ -16,20 +35,23 @@ export class App extends React.Component<{}, State> {
...
+                    value={this.state.query}
+                    onChange={this.handleChange}
+                  />
+                </div>
+              </label>
             </div>
           </div>
 
-          <MoviesList movies={moviesFromServer} />
+          <MoviesList movies={this.visibleMovies()} />
```
Elena-brv, line 54: not do like this, pass to props already calculated movies (call this method in render)

Elena-brv (changes requested):

PR #678, opened 2022-02-23
Create react-movies-list-filter

Pavlo-Khashchevskyi:
[DEMO LINK](https://Pavlo-Khashchevskyi.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,11 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -20,12 +22,16 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={(event) => {
+                  setQuery(event.target.value);
+                }}
```
vitaliy-martynyuk, line 28: it's a bad practice to use arrow functions like that. move it to a separate function

vitaliy-martynyuk (changes requested):
well done! only one fix left

varseniuk (approved):

PR #688, opened 2022-04-01
solution react movies list filter

krolich500grn:
[DEMO LINK](https://krolich500grn.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,34 @@
-import React from 'react';
-import './App.scss';
+import React, { ChangeEvent, useEffect, useState } from 'react';
 import { MoviesList } from './components/MoviesList';
+import { Movie } from './types/Movie';
+
 import moviesFromServer from './api/movies.json';
+import './App.scss';
+
+const singleCaseQuery = (query: string) => {
+  return query.toLowerCase();
+};
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [visibleMovies, setVisibleMovies] = useState<Movie[]>([]);
+
+  useEffect(() => {
+    setVisibleMovies(moviesFromServer);
+  }, []);
+
+  const filterList = (event: ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+
+    setVisibleMovies([...moviesFromServer].filter((movie) => {
+      const title = singleCaseQuery(movie.title);
+      const description = singleCaseQuery(movie.description);
+      const searchValue = singleCaseQuery(event.target.value);
+
+      return title.includes(searchValue) || description.includes(searchValue);
+    }));
+  };
+
   return (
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,9 +1,32 @@
...
+  return query.toLowerCase();
+};
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [visibleMovies, setVisibleMovies] = useState<Movie[]>([]);
+
+  useEffect(() => {
+    setVisibleMovies(moviesFromServer);
+  }, []);
```
jstmpelowycz, line 18: Would you say you actually need to use `useEffect` here?

src/App.tsx
```diff
@@ -1,9 +1,32 @@
...
+import { Movie } from './types/Movie';
+
 import moviesFromServer from './api/movies.json';
+import './App.scss';
+
+const singleCaseQuery = (query: string) => {
+  return query.toLowerCase();
+};
 
 export const App: React.FC = () => {
```
jstmpelowycz, line 12: I guess, you still could have applied `memo` here as well

jstmpelowycz (approved):
Alright!
P.S. if you still strive to optimize these components, consider applying `useCallback`, `useMemo` and so on (as we discussed)

src/App.tsx
jstmpelowycz, line 18: You could just put the array into using `useState`

jstmpelowycz (approved):
Alright!
P.S. if you still strive to optimize these components, consider applying useCallback, useMemo and so on (as we discussed)

src/App.tsx
krolich500grn: I know, but I think will be better to upload less information at the start for a faster page render

src/App.tsx
krolich500grn: I will try to do this 👌

src/App.tsx
```diff
@@ -1,9 +1,32 @@
...
+  const [query, setQuery] = useState('');
+  const [visibleMovies, setVisibleMovies] = useState<Movie[]>([]);
+
+  useEffect(() => {
+    setVisibleMovies(moviesFromServer);
```
Ihor-Karpyn, line 17: use moviesFromServer only here

src/App.tsx
```diff
@@ -1,9 +1,32 @@
...
+  }, []);
+
+  const filterList = (value: string) => {
+    setVisibleMovies(moviesFromServer.filter((movie) => {
```
Ihor-Karpyn, line 21: Don't set in state filtered movies in state. Create function to filter. Write results in variables and then use it. Also use `useMemo` for this vara

Ihor-Karpyn (changes requested):

PR #778, opened 2022-06-06
solutionI

EnjSadman:
[DEMO LINK](https://EnjSadman.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,25 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+function filterMovies(filterValue : string, list : Movie[]) {
+  return (list.filter((movie : Movie) => {
+    const loweredTitle = movie.title.toLowerCase();
+    const loweredDescription = movie.title.toLowerCase();
+
+    if ((loweredTitle.search(filterValue) !== -1)
+    || (loweredDescription.search(filterValue) !== -1)) {
+      return movie;
+    }
+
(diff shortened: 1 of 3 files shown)


ZUBOV-ILLIA (changes requested):
search by description does not work :-(

src/App.tsx
```diff
@@ -1,9 +1,25 @@
...
 
+function filterMovies(filterValue : string, list : Movie[]) {
+  return (list.filter((movie : Movie) => {
+    const loweredTitle = movie.title.toLowerCase();
+    const loweredDescription = movie.description.toLowerCase();
+
+    if ((loweredTitle.search(filterValue) !== -1)
+    || (loweredDescription.search(filterValue) !== -1)) {
+      return movie;
+    }
```
Denchepornyuk, line 14: it is better to use includes method here and remove checking with `-1`

Denchepornyuk, line 14: It is better to make this logic inside MoviesList component

src/App.tsx
```diff
@@ -20,12 +36,19 @@ export const App: React.FC = () => {
...
+                  setQuery(event.target.value);
+                }}
               />
             </div>
           </div>
         </div>
+        {
+          (query === '')
+            ? <MoviesList movies={moviesFromServer} />
+            : <MoviesList movies={filterMovies(query, moviesFromServer)} />
```
Denchepornyuk, line 49: I believe you can remove the ternary operator and always render component with filterMovies

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #785, opened 2022-06-06
add search

Andriyk0:
[DEMO LINK](https://Andriyk0.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,17 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = () => (
+    [...moviesFromServer]
+      .filter(item => item.title.toLowerCase().includes(query.toLowerCase())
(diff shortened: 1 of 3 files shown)


alexandra-protyanova (approved):

src/App.tsx
```diff
@@ -1,9 +1,17 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = () => (
+    [...moviesFromServer]
+      .filter(item => item.title.toLowerCase().includes(query.toLowerCase())
+      || item.description.toLowerCase().includes(query.toLowerCase()))
+  );
```
Denchepornyuk, line 13: It is better to make this logic inside MoviesList component

src/App.tsx
```diff
@@ -18,14 +26,19 @@ export const App: React.FC = () => {
...
+                onChange={(event) => (
+                  setQuery(event.target.value)
+                )}
               />
             </div>
           </div>
         </div>
 
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={visibleMovies()} />
```
Denchepornyuk, line 41: call function `visibleMovies` above return and use result here

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #788, opened 2022-06-07
FilteredMovies was done

vlasiuk-anatolii:
[DEMO LINK](https://vlasiuk-anatolii.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,19 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  function visibleMovies(movies: Movie[], queryy: string) {
+    const filteredMovies = movies.filter(movie => (
+      movie.title.includes(query) || movie.description.includes(queryy)
+    ));
+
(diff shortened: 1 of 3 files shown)


dellachaise (approved):

src/App.tsx
```diff
@@ -1,9 +1,20 @@
...
+  const [query, setQuery] = useState('');
+
+  function visibleMovies(movies: Movie[], queryy: string) {
+    const filteredMovies = movies.filter(movie => (
+      movie.title.toLowerCase().includes(queryy)
+      || movie.description.toLowerCase().includes(queryy)
+    ));
+
+    return filteredMovies;
+  }
```
Denchepornyuk, line 16: It is better to make this logic inside MoviesList component

src/App.tsx
```diff
@@ -20,12 +31,16 @@ export const App: React.FC = () => {
...
+                onChange={(event) => {
+                  setQuery(() => event.target.value);
+                }}
               />
             </div>
           </div>
         </div>
 
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={visibleMovies(moviesFromServer, query)} />
```
Denchepornyuk, line 43: it is better to call visibleMovies function above return and here use result

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #790, opened 2022-06-07
movies-list-filter

anastasiia-nikita:
[DEMO LINK](https://anastasiia-nikita.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,23 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,9 +1,23 @@
...
+    movie.title.toLowerCase().includes(lowerQuery)
+    || movie.description.toLowerCase().includes(lowerQuery)
+  ));
+
+  return filteredMovie;
+}
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = getVisibleMovie([...moviesFromServer], query);
```
iwanttosmile4u, line 19: I think in this case you don't need to copy array with movies, because your source array don't mutate.

Denchepornyuk (approved):

PR #797, opened 2022-06-08
Done

khozhainov-aleksandr:
[DEMO LINK](https://khozhainov-aleksandr.github.io/react_movies-list-filter/)


Yaroslav-Kononenko:
Good work, I did not see any way to improve the code)

src/App.tsx
```diff
@@ -1,9 +1,18 @@
...
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies: Movie[] = moviesFromServer.filter((movie) => {
+    const titleAndDescription = `${movie.title} ${movie.description}`;
+
+    return titleAndDescription.toLowerCase().includes(query.toLowerCase());
```
Anna-Tkachenko, line 13: Now you set query to lower case in each callback call
Will be better to create separate variable before the filter method where you can store query in lower case and use it here

Anna-Tkachenko (approved):

PR #802, opened 2022-06-09
First_commit

igoree88:
[DEMO LINK](https://igoree88.github.io/react_movies-list-filter/)

### DIFF
--- a/src/components/MoviesList/MoviesList.tsx
+++ b/src/components/MoviesList/MoviesList.tsx
@@ -4,12 +4,26 @@ import { MovieCard } from '../MovieCard';
 
 interface Props {
   movies: Movie[];
+  byQuery: string;
 }
 
-export const MoviesList: React.FC<Props> = ({ movies }) => (
-  <div className="movies">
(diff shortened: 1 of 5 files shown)


src/components/MoviesList/MoviesList.tsx
```diff
@@ -4,12 +4,26 @@ import { MovieCard } from '../MovieCard';
...
-    {movies.map(movie => (
-      <MovieCard key={movie.imdbId} movie={movie} />
-    ))}
-  </div>
-);
+export const MoviesList: React.FC<Props> = ({ movies, byQuery }) => {
+  const visibleMovies = movies.filter((movie) => {
+    const lowerTitle = movie.title.toLowerCase();
+    const lowerDescription = movie.description.toLowerCase();
+    const lowerByQuery = byQuery.toLowerCase();
```
Anna-Tkachenko, line 14: Will be better to move this code out of filter method
Now you do the same functionality for each callback call

Anna-Tkachenko (approved):

PR #807, opened 2022-06-13
Solution

Danylo-Dziuban:
[DEMO LINK](https://Danylo-Dziuban.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,12 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -20,16 +23,20 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={
+                  (event) => {
+                    setQuery(event.target.value.toLowerCase());
```
Anna-Tkachenko, line 29: Will be better to transform query to lower case before filtering

Anna-Tkachenko (approved):

PR #825, opened 2022-06-20
solution

huberciak:
[DEMO LINK](https://huberciak.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,11 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -20,12 +22,15 @@ export const App: React.FC = () => {
...
+                onChange={(event) => {
+                  setQuery(event.target.value);
+                }}
               />
             </div>
           </div>
         </div>
 
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={moviesFromServer} query={query} />
```
iLebo-stack, line 33: Instead of passing the query you need to create `visibleMovies` array. Check the `readme` file

iLebo-stack (changes requested):
Good Job 👍, see my comments

src/App.tsx
```diff
@@ -1,9 +1,17 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const filterMovies = moviesFromServer.filter((movie) => (
+    (movie.title || movie.description)
+      .toLowerCase()
+      .includes(query.toLowerCase())
+  ));
```
meforoma, line 13: good one 👍 
better than commonly used approach

meforoma (approved):
great 👍

PR #833, opened 2022-07-10
React movies list filter

vsupruniuk:
[DEMO LINK](https://vsupruniuk.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,10 +10,16 @@ export const App: React.FC = () => {
     setQuery(value);
   };
 
-  const moviesArr = moviesFromServer.filter(movie => {
-    return movie.title.toLowerCase().includes(query.toLowerCase())
-      || movie.description.toLowerCase().includes(query.toLowerCase());
-  });
+  let visibleMovies = moviesFromServer;
+
+  if (query.length) {
+    const queryForFiltration = query.toLowerCase();
+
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,15 +1,25 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleQueryChange = (value: string) => {
+    setQuery(value);
+  };
+
+  const moviesArr = moviesFromServer.filter(movie => {
```
Olexandr6, line 13:
```suggestion
  const visibleMovies = moviesFromServer.filter(movie => {
```

src/App.tsx
```diff
@@ -1,15 +1,25 @@
...
+    return movie.title.toLowerCase().includes(query.toLowerCase())
+      || movie.description.toLowerCase().includes(query.toLowerCase());
+  });
```
Olexandr6, line 16: You use `query.toLowerCase()` twice, let's create a separate variable for this

src/App.tsx
Olexandr6, line 13: Should we filter in case `query` is an empty string?

Olexandr6 (changes requested):
Good job, let's work on improvements

Olexandr6 (approved):

PR #834, opened 2022-07-11
Solution

mira1zu:
- [DEMO LINK](https://mira1zu.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,20 @@
-import React from 'react';
+import React, { useState } from 'react';
+
 import './App.scss';
+
 import { MoviesList } from './components/MoviesList';
+
 import moviesFromServer from './api/movies.json';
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,20 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter((movie) => {
+    const regexp = new RegExp(query, 'i');
+
+    return regexp.test(movie.title) || regexp.test(movie.description);
+  });
```
Olexandr6, line 16: Should we filter in case `query` is an empty string?

Olexandr6 (approved):

PR #837, opened 2022-07-12
solution

AndriiNychaiuk:
- [DEMO LINK](https://AndriiNychaiuk.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,15 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, getQuery] = useState('');
+
+  const visibleMovies = [...moviesFromServer]
+    .filter(movie => movie.title.toLowerCase().includes(query.toLowerCase())
+      || movie.description.toLowerCase().includes(query.toLowerCase()));
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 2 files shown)


mflkvsk (approved):

src/App.tsx
```diff
@@ -1,9 +1,15 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, getQuery] = useState('');
+
+  const visibleMovies = [...moviesFromServer]
```
Olexandr6, line 9: You don't need array copy here, filter doesn't mutate arr
```suggestion
  const visibleMovies = moviesFromServer
```

src/App.tsx
```diff
@@ -1,9 +1,15 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, getQuery] = useState('');
```
Olexandr6, line 7:
```suggestion
  const [query, setQuery] = useState('');
```

src/App.tsx
```diff
@@ -1,9 +1,15 @@
...
+
+  const visibleMovies = [...moviesFromServer]
+    .filter(movie => movie.title.toLowerCase().includes(query.toLowerCase())
+      || movie.description.toLowerCase().includes(query.toLowerCase()));
```
Olexandr6, line 11: Here you do `query.toLowerCase()` twice, let's create a var and do this only once

Olexandr6 (changes requested):
Good job, some small things should be fixed

illiarerun (approved):

PR #838, opened 2022-07-13
filter was added

mflkvsk:
[DEMO LINK](https://mflkvsk.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,23 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,23 @@
...
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+  };
+
+  const preparedQuery = query.toLowerCase();
+
+  const visibleMovies = !query
+    ? [...moviesFromServer]
+    : moviesFromServer
```
AndriiNychaiuk, line 17: Great job! The only thing I recomend to change is the way how you make your array visible. 
```query &&  array.filter(...)``` will do the same thing as the ternary operator in this case

AndriiNychaiuk (approved):

src/App.tsx
mykola-koval: agree, filter should only be executed when query is not empty

mykola-koval (approved):

PR #851, opened 2022-07-18
solution

NikitaKarbazin:
[DEMO LINK](https://nikitakarbazin.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,11 +6,16 @@ import moviesFromServer from './api/movies.json';
 export const App: React.FC = () => {
   const [filterName, setFilterName] = useState('');
 
+  const containsString = (text: string) => text
+    .includes(filterName.toLowerCase());
+
   const visibleMovies = moviesFromServer
-    .filter(movie => movie.title.toLowerCase()
-      .includes(filterName.toLowerCase())
-    || movie.description.toLowerCase()
-      .includes(filterName.toLowerCase()));
+    .filter(movie => {
+      const hasTitle = containsString(movie.title.toLowerCase());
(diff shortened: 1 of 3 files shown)


averstiuk (approved):
I like your decision)))

src/App.tsx
```diff
@@ -1,9 +1,23 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [filterName, setFilterName] = useState('');
+
+  const visibleMovies = moviesFromServer
+    .filter(movie => movie.title.toLowerCase()
+      .includes(filterName.toLowerCase())
+    || movie.description.toLowerCase()
+      .includes(filterName.toLowerCase()));
```
jstmpelowycz, line 13: decompose this logic

jstmpelowycz (changes requested):

src/App.tsx
NikitaKarbazin: Create another variables? Don't understand what exactly I need to decompose

jstmpelowycz: creating a var is one of the possible ways to make it more readable and maintainable
why not to create some function-helper? as an example
```
const containsString = (...) => ...

const lowercaseQuery = ...

const visibleMovies = movies.fromServer.filter(({ ... }) => {
  const hasTitle = ...
  const hasDescription = ...

  return hasTitle || hasDescription;
});
```

jstmpelowycz: is it good?

jstmpelowycz (approved):
awesome 🔥

PR #858, opened 2022-07-18
searchNwatch

Samaev:
[DEMO LINK](https://Samaev.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,15 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+const [query, setQuery] = useState('');
+const moviesToShow = moviesFromServer.filter(movie => {
+  return movie.title.toLowerCase().includes(query.toLowerCase())
+  || movie.description.toLowerCase().includes(query.toLowerCase())
+})
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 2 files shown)


Bohdan-Pyschchyk (approved):

src/App.tsx
```diff
@@ -1,9 +1,15 @@
...
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+const [query, setQuery] = useState('');
+const moviesToShow = moviesFromServer.filter(movie => {
+  return movie.title.toLowerCase().includes(query.toLowerCase())
+  || movie.description.toLowerCase().includes(query.toLowerCase())
```
illiarerun, line 10: let's set query value in `setQuery` with `.toLowerCase()` so you  wouldn't need to do it twice here
```suggestion
    movie.title.toLowerCase().includes(query)
    || movie.description.toLowerCase().includes(query)
```

src/App.tsx
```diff
@@ -19,13 +25,13 @@ export const App: React.FC = () => {
                 type="text"
                 id="search-query"
                 className="input"
-                placeholder="Type search word"
+                onChange={(event) => setQuery(event.target.value)}
```
illiarerun, line 28: create a handler function for it

illiarerun (changes requested):

src/App.tsx
Samaev: ok

src/App.tsx
Samaev: done

src/App.tsx
```diff
@@ -1,9 +1,22 @@
...
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const lowerQuery = query.toLowerCase();
+  
+  const moviesToShow = moviesFromServer.filter(movie => {
+    return movie.title.toLowerCase().includes(lowerQuery)
+  || movie.description.toLowerCase().includes(lowerQuery);
+  });
```
jstmpelowycz, line 14: still looks complicated :(
DRY 
use destructuring, vars – everything that could make it look simpler

Samaev: repaipred
more symbols, but think looks clearer& simpler)

illiarerun (approved):

PR #867, opened 2022-07-18
Solution

Dmytrit:


src/App.tsx
```diff
@@ -1,9 +1,18 @@
...
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const searchedQuery = query.toLowerCase();
+
+  const visibleMovies = (query.length > 0) ? moviesFromServer
+    .filter(movie => movie.title.toLowerCase().includes(searchedQuery)
+    || movie.description.toLowerCase().includes(searchedQuery))
+    : moviesFromServer;
```
jstmpelowycz, line 14: decompose this logic

jstmpelowycz (changes requested):

src/App.tsx
```diff
@@ -1,9 +1,20 @@
...
+  const [query, setQuery] = useState('');
+
+  const searchedQuery = query.toLowerCase();
+
+  const visibleMovies = (query.length > 0) ? moviesFromServer
+    .filter(
+      ({ title, description }) => title.toLowerCase().includes(searchedQuery)
+    || description.toLowerCase().includes(searchedQuery),
+    )
+    : moviesFromServer;
```
jstmpelowycz, line 16: still looks complicated :(
DRY 
here is how I'd implement it
```
const visibleMovies = moviesFromServer.filter(({ ... }) => {
  const hasTitle = ...
  const hasDescription = ...

  return hasTitle || hasDescription;
});
```

jstmpelowycz (changes requested):

illiarerun (approved):

PR #874, opened 2022-07-19
Solution

AntonYegorov:
[DEMO LINK](https://AntonegorovY.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,26 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const queryChange = (value: string) => {
+    setQuery(value);
+  };
+
+  let visibleMovies = moviesFromServer;
+
+  if (query.length) {
+    const queryForFiltration = query.toLowerCase();
+
+    visibleMovies = moviesFromServer.filter(movie => {
+      return movie.title.toLowerCase().includes(queryForFiltration)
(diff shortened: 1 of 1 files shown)


k0nstrukt-or (approved):
:+1:

src/App.tsx
```diff
@@ -1,9 +1,26 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const queryChange = (value: string) => {
+    setQuery(value);
+  };
```
illiarerun, line 11: Take event as a prop, so you can pass handler easier, also you can covert it to `lowerCase` here
```suggestion
  const queryChange = (event) => {
    setQuery(event.target.value.toLowerCase());
  };
```

src/App.tsx
```diff
@@ -1,9 +1,26 @@
...
+
+  let visibleMovies = moviesFromServer;
+
+  if (query.length) {
+    const queryForFiltration = query.toLowerCase();
```
illiarerun, line 16: so you wouldn't need this variable

src/App.tsx
```diff
@@ -1,9 +1,26 @@
...
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const queryChange = (value: string) => {
+    setQuery(value);
+  };
+
+  let visibleMovies = moviesFromServer;
+
+  if (query.length) {
```
illiarerun, line 15: let's turn it into ternary operator, try not to reassign variables
```suggestion
  const visibleMovies = query.length > 0
    ? filter here
    : moviesFromServer
```

illiarerun (changes requested):

PR #878, opened 2022-07-19
finished movies filter in query

omelika:
[DEMO LINK](https://omelika.github.io/react_movies-list-filter/)


src/App.tsx
```diff
@@ -20,12 +29,15 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                onChange={(event) => {
+                  setSearch(event.target.value.trim());
```
LesyaSvk, line 33: Better to create a function handler for it

src/App.tsx
```diff
@@ -1,9 +1,18 @@
...
 
 export const App: React.FC = () => {
+  const [query, setSearch] = useState('');
+
+  const visibleMovies = (moviesFromServer.filter((movie) => {
+    const { title, description } = movie;
+
+    return title.toLowerCase().includes(query.toLowerCase())
+      || description.toLowerCase().includes(query.toLowerCase());
+  }));
```
LesyaSvk, line 14: You can create separate value for ```includes(query.toLowerCase()```. Don't repeat yourselfe

LesyaSvk (approved):
Looks nice)

src/App.tsx
jstmpelowycz: it's not that critical, but indeed makes sense – avoid repeating yourself
proper decomposition is everything

src/App.tsx
jstmpelowycz: agree

jstmpelowycz (changes requested):

src/App.tsx
jstmpelowycz: @LesyaSvk don't be afraid to reject when there are meaningful issues

mykola-koval (approved):

PR #879, opened 2022-07-19
movies-list-filter

dima-pyvovarchuk:
- [DEMO LINK](https://dima-pyvovarchuk.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,16 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    return movie.title.toLowerCase().includes(query.toLowerCase())
+        || movie.description.toLowerCase().includes(query.toLowerCase());
+  });
+
   return (
     <div className="page">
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,16 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    return movie.title.toLowerCase().includes(query.toLowerCase())
+        || movie.description.toLowerCase().includes(query.toLowerCase());
```
illiarerun, line 11: let's set query value in `setQuery` with `.toLowerCase()` so you  wouldn't need to do it twice here, or create a variable with lowerCased query
```suggestion
    return movie.title.toLowerCase().includes(query)
      || movie.description.toLowerCase().includes(query)
```

src/App.tsx
```diff
@@ -20,12 +27,16 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={(event) => {
+                  setQuery(event.target.value);
+                }}
```
illiarerun, line 33: looks like you need to create handler function

illiarerun (changes requested):

illiarerun (approved):

PR #881, opened 2022-07-19
Solution

yanaost:
[DEMO LINK](https://yanaost.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,28 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [visibleMovies, setVisibleMovies] = useState<Movie[]>([]);
+
+  useEffect(() => {
+    setVisibleMovies(
+      moviesFromServer.filter((movie) => {
+        return (
+          movie.title.toLowerCase().includes(query.toLowerCase())
+          || movie.description.toLowerCase().includes(query.toLowerCase())
+        );
+      }),
+    );
+  }, [query]);
+
   return (
     <div className="page">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -20,12 +33,14 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={event => setQuery(event.target.value)}
```
LesyaSvk, line 37: It's better to create hendler for that

src/App.tsx
```diff
@@ -1,15 +1,28 @@
...
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [visibleMovies, setVisibleMovies] = useState<Movie[]>([]);
+
+  useEffect(() => {
+    setVisibleMovies(
+      moviesFromServer.filter((movie) => {
+        return (
+          movie.title.toLowerCase().includes(query.toLowerCase())
+          || movie.description.toLowerCase().includes(query.toLowerCase())
```
LesyaSvk, line 15: You can create separate variable for ```includes(query.toLowerCase()```. Don't repeate yourselfe

LesyaSvk (approved):
Looks nice)

src/App.tsx
illiarerun: agree, but it'd be better to set `lowerCase` value in your handler right away

src/App.tsx
illiarerun: agree, create handler

illiarerun (changes requested):

src/App.tsx
yanaost: we use it once, why do we need to create separate handler?

src/App.tsx
yanaost: I dont understand

jstmpelowycz (approved):

PR #882, opened 2022-07-19
solution

Victance:
[DEMO LINK](https://victance.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,28 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  let visibleMovies;
+
+  if (moviesFromServer.find(movie => (
+    movie.title.toLocaleLowerCase().includes(query.toLocaleLowerCase())
+  ))) {
+    visibleMovies = moviesFromServer.filter(movie => (
+      movie.title.toLocaleLowerCase().includes(query.toLocaleLowerCase())
+    ));
+  }
+
+  if (moviesFromServer.find(movie => (
+    movie.description.toLocaleLowerCase().includes(query.toLocaleLowerCase())
+  ))) {
+    visibleMovies = moviesFromServer.filter(movie => (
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -20,12 +39,15 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                onChange={(event) => {
+                  setQuery(event.target.value);
+                }}
```
illiarerun, line 44: create a function handler for that

src/App.tsx
```diff
@@ -1,9 +1,28 @@
...
+    ));
+  }
+
+  if (moviesFromServer.find(movie => (
+    movie.description.toLocaleLowerCase().includes(query.toLocaleLowerCase())
+  ))) {
+    visibleMovies = moviesFromServer.filter(movie => (
+      movie.description.toLocaleLowerCase().includes(query.toLocaleLowerCase())
+    ));
+  }
```
illiarerun, line 24: too much going on here:
- you don't need this checks before filtering, you just go through array 2 times
- you can combine both conditions in 1 filter function
- you repeat yourself with a `query.toLocaleLowerCase()` a lot

src/App.tsx
illiarerun: and also, you can set it with `.toLowerCase` right away

illiarerun (changes requested):

src/App.tsx
```diff
@@ -1,9 +1,20 @@
...
+  const [query, setQuery] = useState('');
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value.toLowerCase());
+  };
+
+  const visibleMovies = moviesFromServer.filter(movie => (
+    movie.title.toLocaleLowerCase().includes(query)
+    || movie.description.toLocaleLowerCase().includes(query)
+  ));
```
jstmpelowycz, line 16: still looks complicated, but okay
here you could have used destructuring, vars, new separate function like `containsQuery` – everything that could make it look simpler

jstmpelowycz (approved):

jstmpelowycz (changes requested):
you need a small fix

illiarerun (approved):
GJ

PR #883, opened 2022-07-19
Solution created

MariiaNH:
[DEMO LINK](https://mariianh.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,14 @@
-import React from 'react';
+import { FC, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-export const App: React.FC = () => {
+export const App: FC = () => {
+  const [search, setSearch] = useState('');
+  const visibleMovies = moviesFromServer
+    .filter(movie => movie.title.toLowerCase().includes(search)
+    || movie.description.toLowerCase().includes(search));
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -20,12 +25,15 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                onChange={
+                  ({ target }) => setSearch(() => target.value.toLowerCase())
+                }
```
garikr, line 30: For me destructuring and manipulations width data in layout complicates reading
(event) => setSearch(event.target.value) is easier 
and you can format your input in filter. Try it, may be you'll like it too )

garikr (approved):
Gj! But have 1 suggestion

src/App.tsx
jstmpelowycz: too much for a single callback, you'd rather create a separate function-handler
agree, such a structure like this is too much
```
onChange={
  callbackA => function(callbackB)
}

```

src/App.tsx
```diff
@@ -1,9 +1,14 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-export const App: React.FC = () => {
+export const App: FC = () => {
+  const [search, setSearch] = useState('');
+  const visibleMovies = moviesFromServer
+    .filter(movie => movie.title.toLowerCase().includes(search)
+    || movie.description.toLowerCase().includes(search));
```
jstmpelowycz, line 10: decompose this logic

jstmpelowycz (changes requested):

illiarerun (approved):

PR #907, opened 2022-08-10
Develop

LordAsmodey:
[DEMO LINK](https://LordAsmodey.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,8 +5,9 @@ import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
   const [searchQuery, setSearchQuery] = useState('');
-  const sortedMoviesList = moviesFromServer.filter(movie => {
-    return movie.title.toLowerCase().includes(searchQuery.toLowerCase());
+  const visibleMovies = moviesFromServer.filter(movie => {
+    return movie.title.toLowerCase().includes(searchQuery.toLowerCase())
+      || movie.description.toLowerCase().includes(searchQuery.toLowerCase());
   });
 
   return (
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,15 +1,20 @@
...
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [searchQuery, setSearchQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(movie => {
+    return movie.title.toLowerCase().includes(searchQuery.toLowerCase())
+      || movie.description.toLowerCase().includes(searchQuery.toLowerCase());
```
Kras4iy, line 10: i think it is better to do the lowercase method one time instead two ( setSearchQuery(blablabla.toLowerCase()) )

Kras4iy (approved):

src/App.tsx
Kras4iy: also user can missclick and put two or more spaces after input text and there will be no result((

Ihor-Karpyn, line 10: move the filter to a separate function. hard to understanding

Ihor-Karpyn (changes requested):

src/App.tsx
```diff
@@ -1,15 +1,20 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [searchQuery, setSearchQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(movie => {
```
Ihor-Karpyn, line 8: add empty line

Ihor-Karpyn (approved):

PR #914, opened 2022-08-11
Solution

Lentyaeev:
[DEMO LINK](https://lentyaeev.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,36 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  useEffect(() => {
+    const list = document
+      .querySelector('.wrapper')?.children[0].children;
+
+    if (list !== undefined) {
+      for (let i = 0; i < list.length; i += 1) {
+        list[i].classList.add('animate__animated');
+        list[i].classList.add('animate__flash');
+      }
+
+      setTimeout(() => {
+        for (let i = 0; i < list.length; i += 1) {
+          list[i].classList.remove('animate__animated');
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,36 @@
...
+      }
+
+      setTimeout(() => {
+        for (let i = 0; i < list.length; i += 1) {
+          list[i].classList.remove('animate__animated');
+          list[i].classList.remove('animate__flash');
+        }
+      }, 500);
+    }
+  });
```
AleksandrYuschenko, line 26: Круто виглядає. Гарно але шось я не догнав)))

AleksandrYuschenko (approved):
🥇

src/App.tsx
```diff
@@ -1,9 +1,36 @@
...
+        list[i].classList.add('animate__flash');
+      }
+
+      setTimeout(() => {
+        for (let i = 0; i < list.length; i += 1) {
+          list[i].classList.remove('animate__animated');
+          list[i].classList.remove('animate__flash');
+        }
+      }, 500);
+    }
```
solaryasha, line 25: Let's do it with state like
`const [isLoadiing, setIsLoading] = useState(true)`
and use classname for passing these classes


In setTimeout you just set loading to be `false`

solaryasha (changes requested):

src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    setIsLoading(true);
+    setTimeout(() => setIsLoading(false), 300);
```
solaryasha, line 12: You should clearTimeout function with setTimeout id

solaryasha (changes requested):

solaryasha (approved):

PR #915, opened 2022-08-11
implemented filter for movie list

nikita-biriukov:
[DEMO LINK](https://nikita-biriukov.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,18 +4,11 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
-  const [query, setQuery] = useState<string>('');
-
-  const handleSearch = (event: React.ChangeEvent<HTMLInputElement>) => {
-    setQuery(event.target.value);
-  };
+  const [query, setQuery] = useState('');
 
   const visibleMovies = moviesFromServer.filter(movie => {
-    const { title, description } = movie;
-    const lowerCaseQuery = query.toLowerCase();
-
-    return title.toLowerCase().includes(lowerCaseQuery)
-    || description.toLowerCase().includes(lowerCaseQuery);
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,23 @@
...
+  const handleSearch = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+  };
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const { title, description } = movie;
+    const lowerCaseQuery = query.toLowerCase();
+
+    return title.toLowerCase().includes(lowerCaseQuery)
+    || description.toLowerCase().includes(lowerCaseQuery);
```
denys-danyliuk, line 18: maybe this is kinda overkill, you will create this variables on each iteration, maybe I'm wrong and this how it should be

denys-danyliuk (approved):
Good job🔥

src/App.tsx
```diff
@@ -1,9 +1,16 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    return movie.title.toLowerCase().includes(query.toLowerCase())
+    || movie.description.toLowerCase().includes(query.toLowerCase());
```
ivan-kadykalo, line 11: Good work, but it will be more readable if you will create separate variables, and compare them, because at now it looks hard)) 
Shortest is not the best in every time)

```suggestion
    const titleInclude = movie.title.toLowerCase().includes(query.toLowerCase());
    const decriptionInclude = movie.description.toLowerCase().includes(query.toLowerCase());
    
    return titleInclude || decriptionInclude;
```

ivan-kadykalo (approved):

PR #936, opened 2022-08-14
task is ready

opanasiukvlad:
[DEMO](https://opanaskiukvlad.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,23 @@
-import React from 'react';
+import React, {useState} from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const filteredmoviesFromServer = moviesFromServer.filter((movie) => {
+    if (movie.title.toLowerCase().includes(query.toLowerCase())) {
+      return movie;
+    }
+
+    if (movie.description.toLowerCase().includes(query.toLowerCase())) {
+      return movie;
+    }
+
+    return 0;
+  });
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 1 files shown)


ClaudiaKalnitskaya (approved):
Good!

src/App.tsx
```diff
@@ -1,9 +1,23 @@
...
+  const [query, setQuery] = useState('');
+
+  const filteredmoviesFromServer = moviesFromServer.filter((movie) => {
+    if (movie.title.toLowerCase().includes(query.toLowerCase())) {
+      return movie;
+    }
+
+    if (movie.description.toLowerCase().includes(query.toLowerCase())) {
+      return movie;
+    }
```
solaryasha, line 16: Don't repeat yourself

src/App.tsx
```diff
@@ -20,12 +34,16 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={(event) => {
+                  setQuery(event.target.value);
+                }}
```
solaryasha, line 40: Let's avoid passing inline function, let's create separate function for this

solaryasha (changes requested):

src/App.tsx
opanasiukvlad: Thanks, fixed!

src/App.tsx
```diff
@@ -1,9 +1,16 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const filteredMoviesFromServer = moviesFromServer.filter((movie) => {
+    return movie.title.toLowerCase().includes(query.toLowerCase())
+    || movie.description.toLowerCase().includes(query.toLowerCase());
```
solaryasha, line 11: You could move the result of toLowerCase method to the separate variable

solaryasha (approved):

PR #961, opened 2022-09-08
Solution

Dmytro-Matsehora:
[DEMO LINK](https://Dmytro-Matsehora.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,12 +4,12 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
-  const [query, filtered] = useState('');
+  const [query, setQuery] = useState('');
 
   const handleFilter = (event: React.ChangeEvent<HTMLInputElement>) => {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,21 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleFilter = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+
+    setQuery(value);
+  };
```
tiserett, line 13:
```suggestion
  const handleFilter = (event: React.ChangeEvent<HTMLInputElement>) => {
    const { value } = event.target;

    setQuery(value);
  };
```
you don't really need this piece of code, on the onChange take event in props, and then call setQuery(event.targey.value)

tiserett (changes requested):
you should insert few changes

tiserett (approved):
🔥🔥🔥

PR #964, opened 2022-09-09
solution

Githi54:
[DEMO LINK](https://githi54.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,29 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer
+    .filter(({ title, description }) => (
+      title
+        .toLowerCase()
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,15 +1,29 @@
...
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer
+    .filter(({ title, description }) => (
+      title
+        .toLowerCase()
+        .includes(query.toLowerCase())
+      || description
+        .toLowerCase()
+        .includes(query.toLowerCase())
+    ));
```
AndreyJigijalkin, line 16: You can make small separate function to compare two string values in lowerCase :)

AndreyJigijalkin (approved):
Very good:)

src/App.tsx
Githi54: Why?)

dellachaise: Because you kind of duplicate your logic twice(do the same but once on `title`, then on `description`. And better use function for it.

dellachaise (approved):
Well done 👍 

Just better do not repeat yourself(comments) - [checklist point 2](https://github.com/mate-academy/react_movies-list-filter/blob/master/checklist.md)

PR #967, opened 2022-09-14
develop

Blankevich:
https://Blankevich.github.io/react_movies-list-filter/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,19 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const getVisibleMovies = () => {
+    return moviesFromServer.filter(movie => movie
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,19 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const getVisibleMovies = () => {
+    return moviesFromServer.filter(movie => movie
+      .title.toLowerCase().includes(query.toLocaleLowerCase())
+      || movie.description.toLowerCase().includes(query.toLocaleLowerCase()));
+  };
```
stanislav-topikha, line 13:
```suggestion
  const getVisibleMovies = () => {
    return moviesFromServer.filter(movie => 
      stringContainsCaseInsensitive(movie.title, query)
      || stringContainsCaseInsensitive(movie.description, query)
  };
```
move part of the logic to the function

stanislav-topikha (changes requested):

src/App.tsx
Blankevich: done

TarasHoliuk (approved):
Well done :+1: Approved

PR #973, opened 2022-09-16
solution

oksana-logos-frontend:
[DEMO LINK](https://oksana-logos-frontend.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,19 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer
+    .filter(({ title, description }) => {
+      return (
+        title.toLowerCase().includes(query.toLocaleLowerCase())
+        || description.toLowerCase().includes(query.toLocaleLowerCase())
+      );
+    });
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,19 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer
+    .filter(({ title, description }) => {
+      return (
+        title.toLowerCase().includes(query.toLocaleLowerCase())
+        || description.toLowerCase().includes(query.toLocaleLowerCase())
```
VolodymyrChuyko, line 13: You can concatenate a title and description to simplify your code. Also, you probably don't want to filter an array if there is no query.

src/App.tsx
```diff
@@ -20,12 +30,17 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                name="query"
+                value={query}
+                onChange={(event) => (
+                  setQuery(event.target.value)
```
VolodymyrChuyko, line 36: Here is some [info](https://www.anycodings.com/1questions/4778360/when-to-use-inline-function-on-button-onclick-event-javascriptreactjs) on when it's better to use a handler than an inline function.

VolodymyrChuyko (changes requested):
Great job! Please, pay attention to the comments

src/App.tsx
oksana-logos-frontend: Done!

VolodymyrChuyko (approved):
Looks good to me!

AlinaMatuschak (approved):

PR #979, opened 2022-09-18
react_movies-list-filter

yurii-kozlov:
Demo link - https://yurii-kozlov.github.io/react_movies-list-filter/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,20 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleSearch = (value: string) => setQuery(value);
+  const lowerText = (str: string) => {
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,20 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleSearch = (value: string) => setQuery(value);
```
SanyaBratashchuk, line 9: How this function help you?

SanyaBratashchuk (changes requested):

src/App.tsx
yurii-kozlov: > How this function help you?

@SanyaBratashchuk it's just for convenience. Should I delete it and make all changes via "onChange" directly?

SanyaBratashchuk: It is better to create handler which accept an event and set its target's value to the state

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #992, opened 2022-09-20
React solution: Movies List Filter

wfurs0w:
[DEMO LINK](https://wfurs0w.github.io/react_movies-list-filter/)


src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(({ title, description }) => {
+    const lowerCaseCompare = query.toLowerCase();
```
rboiar, line 10: Optionally: you can make it a function and compare `query` with `title` and `description` here

And use it below to avoid code duplication

```suggestion
   const lowerCaseCompare = (string: string) => (
    string.toLowerCase().includes(query.toLowerCase())
  );
```

rboiar (approved):
Good job!

src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
+
+    return (
+      title.toLowerCase().includes(lowerCaseCompare)
+      || description.toLowerCase().includes(lowerCaseCompare)
```
AndriiRut, line 14: if you wrote the same code twice you need to create a function for this code and write it down before the component if it doesn't receive any state variables.

AndriiRut (changes requested):

AndriiRut (approved):
🚀

PR #1031, opened 2022-10-01
Done

Oleksii-Blyzniuk:
https://Oleksii-Blyzniuk.github.io/react_movies-list-filter/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,20 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -18,14 +29,16 @@ export const App: React.FC = () => {
               <input
                 type="text"
                 id="search-query"
+                value={value}
+                onChange={(event) => setValue(event.target.value)}
```
liliia-livandovskaya, line 33: that is not a mistake but its better to add a handler there

liliia-livandovskaya (approved):
Good job

PR #1040, opened 2022-10-10
Develop

11Olya112:


polosanya (changes requested):
Your demo link doesn't work. Please fix it

src/App.tsx
```diff
@@ -1,9 +1,19 @@
...
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const getvisibleMovies = () => {
+    return moviesFromServer.filter(movie => movie.title.toLowerCase()
+      .includes(query.toLowerCase()) || movie.description.toLowerCase()
+      .includes(query.toLowerCase()));
+  };
+
+  const visibleMovies = getvisibleMovies();
```
Anastasiia-Svintsova, line 15: You don't need a separate function here.
```suggestion
  const visibleMovies = moviesFromServer.filter(movie => 
    movie.title.toLowerCase() .includes(query.toLowerCase()) || 
    movie.description.toLowerCase().includes(query.toLowerCase()));
  };
```

Anastasiia-Svintsova (changes requested):
Good! Just fix something

polosanya (changes requested):
It is only App.tsx in your request. Please add all other files while committing, so we can check them.
image

src/App.tsx
```diff
@@ -1,9 +1,16 @@
...
+import { MoviesList } from './components/MoviesList';
+import './App.scss';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer
+    .filter(movie => movie.title.toLowerCase()
+      .includes(query.toLowerCase()) || movie.description.toLowerCase()
+      .includes(query.toLowerCase()));
```
Bohdan-Maksymiuk, line 12: move `query.toLowerCase()` to separate variable instead of recalculate it twice on every iteration
then use this variable here

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #1044, opened 2022-10-17
Solution

G0odvin:
[DEMO LINK](https://G0odvin.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,24 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter((item) => {
+    if (query === '') {
+      return item;
+    }
+
+    if (item.title.toLowerCase().includes(query.toLowerCase())
+    || item.description.toLowerCase().includes(query.toLowerCase())) {
+      return item;
+    }
+
+    return false;
+  });
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter((item) => {
+    if (query === '') {
+      return item;
+    }
```
Anastasiia-Svintsova, line 12: actually, you don't need this check

src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
+
+    if (item.title.toLowerCase().includes(query.toLowerCase())
+    || item.description.toLowerCase().includes(query.toLowerCase())) {
+      return item;
+    }
```
Anastasiia-Svintsova, line 17: Just return it. It will turn into true or fall, what is needed for the filter method. And optional move query.toLowerCase() to a separate const
```suggestion
    return  (
     item.title.toLowerCase().includes(query.toLowerCase())
      || item.description.toLowerCase().includes(query.toLowerCase())
    )
```

Anastasiia-Svintsova (changes requested):
Fix comments please

src/App.tsx
```diff
@@ -1,9 +1,19 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const queryToLowerCase = query.toLowerCase();
+
+  const visibleMovies = moviesFromServer.filter((item) => {
+    return (
```
Bohdan-Maksymiuk, line 11:
```suggestion
  const visibleMovies = moviesFromServer.filter((item) => (
```
you may use shorthand return here

Bohdan-Maksymiuk (approved):
Good job 👍

PR #1051, opened 2022-10-24
solution

vmamchur:
[DEMO LINK](https://vmamchur.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,20 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
-import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+import { MoviesList } from './components/MoviesList';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = [...moviesFromServer].filter(movie => {
+    const titleToLower = movie.title.toLowerCase();
+    const descriptionToLower = movie.description.toLowerCase();
+    const queryToLower = query.toLowerCase();
+
+    return titleToLower.includes(queryToLower)
+      || descriptionToLower.includes(queryToLower);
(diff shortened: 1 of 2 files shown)


imaksweb (approved):
Awesome 👽

src/App.tsx
```diff
@@ -1,9 +1,20 @@
...
 import moviesFromServer from './api/movies.json';
+import { MoviesList } from './components/MoviesList';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const titleToLower = movie.title.toLowerCase();
+    const descriptionToLower = movie.description.toLowerCase();
+    const queryToLower = query.toLowerCase();
```
naziamoff, line 12: Great job with creating variables for making the code clearer 👍 
You can make it better still. Create boolean conditions for `doesTitleMatch / doesDescriptionMatch` instead of those

naziamoff (changes requested):
Does not work if query is spaced

Screenshot 2022-10-28 at 12 15 53

src/App.tsx
```diff
@@ -1,9 +1,20 @@
...
+import { MoviesList } from './components/MoviesList';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const doesTitleMatch = movie.title.toLowerCase()
+      .includes(query.toLowerCase());
+    const doesDescriptionMatch = movie.description.toLowerCase()
+      .includes(query.toLowerCase());
```
Anna-Tkachenko, line 13: will be better to create separate variable for storing query in lower case

Anna-Tkachenko (approved):

naziamoff (approved):

PR #1052, opened 2022-10-24
solution was added

Maksym845:
- [DEMO LINK](https://Maksym845.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,11 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [searchWord, setSearchWord] = useState('');
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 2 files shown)


Yeliseiev (approved):
An interesting solution. But it seems to me that it is better the already filtered list provide to the component

src/App.tsx
```diff
@@ -20,12 +22,19 @@ export const App: React.FC = () => {
...
               />
             </div>
           </div>
         </div>
 
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList
+          movies={moviesFromServer}
+          searchWord={searchWord}
+        />
```
sergii-nosachenko, line 37: No need to use filter logic inside MovieList component. Also your task says you need to create state variable with `visibleMovies` and pass it here as props.

src/components/MoviesList/MoviesList.tsx
```diff
@@ -4,11 +4,17 @@ import { MovieCard } from '../MovieCard';
...
-export const MoviesList: React.FC<Props> = ({ movies }) => (
+export const MoviesList: React.FC<Props> = ({ movies, searchWord }) => (
   <div className="movies">
-    {movies.map(movie => (
+    {movies.filter((movie) => {
+      return (movie.title.toLowerCase()
+        .includes(searchWord.toLowerCase())
+      || movie.description.toLowerCase()
+        .includes(searchWord.toLowerCase()));
+    }).map(movie => (
```
sergii-nosachenko, line 17: This function belongs to parent component. Also create create constants f.e. `descriptionNormalized` and use it. Additional lines of code but a huge improvement in readability.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Awesome 🔥🔥🔥

PR #1055, opened 2022-10-25
Implemented the search field for movies

anastasiia-nahaiska:
[DEMO LINK](https://anastasiia-nahaiska.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,23 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+const filtrateMovies = (movies: Movie[], query: string) => {
+  return movies.filter(movie => (
+    movie.description.toLowerCase().includes(query.toLowerCase())
+    || movie.title.toLowerCase().includes(query.toLowerCase())
+  ));
+};
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = filtrateMovies(moviesFromServer, query);
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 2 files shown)


Apollonia5 (approved):
GJ!

src/App.tsx
```diff
@@ -1,9 +1,23 @@
...
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+const filtrateMovies = (movies: Movie[], query: string) => {
+  return movies.filter(movie => (
+    movie.description.toLowerCase().includes(query.toLowerCase())
+    || movie.title.toLowerCase().includes(query.toLowerCase())
+  ));
```
naziamoff, line 10: create separate variables for those conditions

src/App.tsx
```diff
@@ -1,9 +1,23 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+const filtrateMovies = (movies: Movie[], query: string) => {
```
naziamoff, line 6: What about this way?
```suggestion
const getMatchingMovies = (movies: Movie[], query: string) => {
```

src/App.tsx
```diff
@@ -1,9 +1,23 @@
...
+  return movies.filter(movie => (
+    movie.description.toLowerCase().includes(query.toLowerCase())
+    || movie.title.toLowerCase().includes(query.toLowerCase())
+  ));
+};
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = filtrateMovies(moviesFromServer, query);
```
naziamoff, line 15: do you really have to pass moviesFrom Server? To me it looks like it is imported and function already has an access to it

naziamoff (changes requested):
You should handle the case if there are spaces around my query. It does not work now

Anna-Tkachenko (approved):
⭐️

PR #1061, opened 2022-10-26
Implement rendering filtered movies list

imaksweb:
- [DEMO LINK](https://imaksweb.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,21 @@
-import React from 'react';
+import { FC, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-export const App: React.FC = () => {
+export const App: FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleInputChange = (newValue: string) => {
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,21 @@
...
+
+  const handleInputChange = (newValue: string) => {
+    setQuery(newValue);
+  };
+
+  const visibleMovies = moviesFromServer
+    .filter(movie => (
+      movie.title.toLowerCase().includes(query.toLowerCase())
+        || movie.description.toLowerCase().includes(query.toLowerCase())
+    ));
```
naziamoff, line 17: any ideas on improving readability?

naziamoff (changes requested):
does not work if spaced
Screenshot 2022-10-28 at 11 51 31

imaksweb:
> does not work if spaced Screenshot 2022-10-28 at 11 51 31

I fixed this problem but tests are crashed now, I just wanted to practice a bit with `debounce` 🙂

Anna-Tkachenko (approved):
⭐️

PR #1066, opened 2022-10-26
Movies list - Filter

Vladosik1337:
[DEMO LINK](https://Vladosik1337.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,29 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(movie => {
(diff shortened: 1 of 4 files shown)


a-nahaitsev (approved):
Good job! 👍

src/App.tsx
```diff
@@ -1,16 +1,29 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const title = movie.title.toLowerCase();
+    const description = movie.description.toLowerCase();
+    const filteredQuery = query.toLowerCase();
```
naziamoff, line 11: we can skip those variables and create only booleans for checking a condition.
Let's say `doesTitleMatch / doesDescriptionMatch`

naziamoff (changes requested):
does not work if query is spaced
Screenshot 2022-10-28 at 11 53 50

naziamoff (approved):
👍

PR #1067, opened 2022-10-26
Filter for movies list has been created

iam-datsenko:
[DEMO LINK](https://iam-datsenko.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+const getFilteredMovies = (movie: Movie, search: string) => {
+  const searchBy = search.toLowerCase();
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,22 @@
...
+
+  return title.includes(searchBy) || description.includes(searchBy);
+};
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer
+    .filter(movies => getFilteredMovies(movies, query));
+
```
iPanda177, line 19: i think better to save movies in the state, and just update state after filtering

iPanda177 (approved):

src/App.tsx
Denchepornyuk: agree, let's save movies in state and remove `visibleMovies` variable

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1072, opened 2022-10-26
Develop

artem-nasadchuk:
[DEMO LINK](https://artem-nasadchuk.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,21 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,21 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const { title, description } = movie;
+    const lowCaseTitle = title.toLowerCase();
+    const lowCaseDescription = description.toLowerCase();
+    const lowCaseQuery = query.toLowerCase();
```
naziamoff, line 13: let's make boolean conditions like `doesTitleMatch / doesDescriptionMatch` instead of those. Will be clearer this way

naziamoff (changes requested):
does not work if query is spaced
Screenshot 2022-10-28 at 12 26 55

Anna-Tkachenko (approved):

PR #1074, opened 2022-10-26
movies list filter implemented

artem-musii:
[DEMO LINK](https://artem-musii.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,19 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const checkPresence = (value: string) => {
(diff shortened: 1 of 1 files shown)


anton-iskryk (changes requested):
The search doesn't work with extra spaces
Screenshot 2022-10-28 at 12 34 32

anton-iskryk (approved):

src/App.tsx
```diff
@@ -1,9 +1,19 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const checkPresence = (value: string) => {
+    return value.toLowerCase().includes(query.toLowerCase());
+  };
```
naziamoff, line 11: great decision to separate this logic from filtering 🔥 
We can improve naming though
```suggestion
  const doesQueryMatch = (value: string) => {
    return value.toLowerCase().includes(query.toLowerCase());
  };
```

naziamoff (approved):
Well done 🔥

PR #1084, opened 2022-10-27
Implement Search movie input with title and description filter

avkovlit:
[DEMO LINK](https://avkovlit.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,24 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = () => {
+    return (
+      moviesFromServer.filter(
+        movie => movie.title.toLowerCase().includes(query.toLowerCase())
+        || movie.description.toLowerCase().includes(query.toLowerCase()),
+      )
+    );
+  };
+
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = () => {
+    return (
+      moviesFromServer.filter(
+        movie => movie.title.toLowerCase().includes(query.toLowerCase())
+        || movie.description.toLowerCase().includes(query.toLowerCase()),
+      )
```
naziamoff, line 14: any ideas on improving its readability?
You can ping me if you don't have any, but I believe you can do it

src/App.tsx
```diff
@@ -1,9 +1,24 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = () => {
```
naziamoff, line 9: If it looks like a duck, swims like a duck, and quacks like a duck, then it should be named a duck.
Function name normally starts with a verb. Those are `get...`, `set...`, `calculate...`, `switch...` and so on

naziamoff (changes requested):
does not work if I space my query
Screenshot 2022-10-28 at 11 46 11

PR #1085, opened 2022-10-27
implemented react-movies-list-filter

nikitaovch99:


src/App.tsx
```diff
@@ -1,9 +1,15 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(
+    (movie => movie.title.toLowerCase().includes(query.toLowerCase())
+    || movie.description.toLowerCase().includes(query.toLowerCase())),
+  );
```
naziamoff, line 11: create separate variables for `doesTitleMatch / doesDescriptionMatch` - it will be more readable this way

naziamoff (changes requested):
does not work if the query is spaced. I mean if I write let's say (`          inception         `)

src/App.tsx
naziamoff: I meant you should create variables but not functions. It looked not bad actually, it just was needed to be adjusted)

naziamoff (approved):

PR #1088, opened 2022-10-27
react_movies-list-filter

xiimiik:
[DEMO LINK](https://xiimiik.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -38,7 +38,11 @@ export const App: React.FC = () => {
           </div>
         </div>
 
-        <MoviesList movies={filterByInput(query)} />
+        <MoviesList
+          movies={filterByInput(
+            query.toLowerCase(),
+          )}
+        />
       </div>
 
       <div className="sidebar">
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,20 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+const filterByInput = (input: string) => {
+  return moviesFromServer.filter(movie => {
+    const title = movie.title.toLowerCase();
+    const description = movie.description.toLowerCase();
+
+    return title.match(input) || description.match(input);
```
DenysKleimenov, line 11: I think the **match** method is not needed here, because the **match()** method retrieves the result of matching a string against a regular expression, but our input is not a regular expression, so **includes()** would be more appropriate.

denyskleimenov (approved):

src/App.tsx
xiimiik: includes might to be better, but matrch returns an Array containing the results of the match, or null if nothing was matched, so result will be same

silchyn (approved):
Cool!

PR #1091, opened 2022-10-27
Implement case insensitive film search

anton-iskryk:
[DEMO LINK](https://anton-iskryk.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const { title, description } = movie;
+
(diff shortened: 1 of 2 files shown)


artem-musii (approved):
GJ

src/App.tsx
```diff
@@ -1,9 +1,22 @@
...
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const { title, description } = movie;
+
+    const lowerCaseTitle = title.toLowerCase();
+    const lowerCaseDescription = description.toLowerCase();
+    const lowerCaseQuery = query.toLowerCase();
```
naziamoff, line 14: Good job, that is really nice thing to create variables to improve readability
You could do it simpler though. 

lets make but 2 conditions though. 
If you do it like `const doesTitleMatch = title.toLowerCase().includes(query.toLowerCase())` it will be even more readable and simple. Return will look more clear too.

naziamoff (changes requested):
doesn't work if I space my query
Screenshot 2022-10-28 at 11 33 39

naziamoff (approved):

PR #1094, opened 2022-10-27
React_movies-list-filter

DenysKleimenov:
[DEMO LINK](https://DenysKleimenov.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,28 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  useEffect(() => {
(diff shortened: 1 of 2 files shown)


xiimiik (approved):
Nice!

src/App.tsx
```diff
@@ -1,9 +1,28 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  useEffect(() => {
+    const lowerQuery = query.toLowerCase();
+
+    setMovies(moviesFromServer.filter(movie => {
```
Anna-Tkachenko, line 13: if you set new state depending on the current one, you should use functional setState

Anna-Tkachenko (changes requested):

src/App.tsx
DenysKleimenov: I set state depending on the data that I got from the server, not the current state.

Anna-Tkachenko: I see, my bad, sorry

Anna-Tkachenko (approved):

PR #1095, opened 2022-10-27
Solved. Ver.1.0

PaulVoron:
[DEMO LINK](https://PaulVoron.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,26 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(
+    (movie) => {
+      if (movie.title.toLowerCase().includes(query.toLowerCase())
+        || (movie.description.toLowerCase().includes(query.toLowerCase()))) {
+        return true;
+      }
+
+      return false;
+    },
+  );
+
   return (
     <div className="page">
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,15 +1,26 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(
```
Anastasiia-Svintsova, line 8:
```suggestion
  const [query, setQuery] = useState('');
  
  const visibleMovies = moviesFromServer.filter(
```

src/App.tsx
```diff
@@ -1,15 +1,26 @@
...
+    (movie) => {
+      if (movie.title.toLowerCase().includes(query.toLowerCase())
+        || (movie.description.toLowerCase().includes(query.toLowerCase()))) {
```
Anastasiia-Svintsova, line 11: Move it to a separate method

Anastasiia-Svintsova (changes requested):

src/App.tsx
PaulVoron: I don't think it will be better solution. Now there are just a 2 lines and this variant is very clear for understanding the code.
I would like insist on this variant. Please approve it.

polosanya: Shorter is not always better.
Sometimes it is better to write extra code but make it as clear as possible, especially if you are working on a real project with a team of developers.

polosanya (approved):

PR #1099, opened 2022-10-27
Develop

vallerii:
[DEMO LINK](https://vallerii.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,24 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
+  imgUrl: string;
+  imdbUrl: string;
+  imdbId: string;
+}
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => (
+    movie.description.toLowerCase().includes(query.toLowerCase())
```
Anna-Tkachenko, line 18: will be better to create separate variable for storing query in lower case

Anna-Tkachenko (approved):

PR #1103, opened 2022-10-28
implemented filter option

ArtemHulkoDev:
[DEMO LINK](https://ArtemHulkoDev.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,19 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, changeQuery] = useState('');
+
+  const filterMovies = (value: string) => {
+    return moviesFromServer
+      .filter(movie => (
+        movie.title.toLowerCase().match(value.toLowerCase())
+        || movie.description.toLowerCase().match(value.toLowerCase())
+      ));
+  };
+
   return (
     <div className="page">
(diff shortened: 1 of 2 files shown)


vallerii:
Good job!

src/App.tsx
```diff
@@ -1,9 +1,19 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, changeQuery] = useState('');
+
+  const filterMovies = (value: string) => {
+    return moviesFromServer
+      .filter(movie => (
+        movie.title.toLowerCase().match(value.toLowerCase())
+        || movie.description.toLowerCase().match(value.toLowerCase())
```
silchyn, line 13: Hard to read

silchyn (changes requested):

src/App.tsx
ArtemHulkoDev: fixed

src/App.tsx
```diff
@@ -1,9 +1,23 @@
...
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, changeQuery] = useState('');
+
+  const isMatching = (prop: string, value: string) => {
+    return prop.toLowerCase().match(value.toLowerCase());
```
Anna-Tkachenko, line 10: will be better to create separate variable for storing query in the lower case before the calling filter method

Anna-Tkachenko (approved):

PR #1106, opened 2022-10-28
implemented React movies list

VovaDzhuryk:
[DEMO LINK](https://VovaDzhuryk.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,24 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './react-app-env';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  useEffect(() => {
+    const lowerQuery = query.toLowerCase();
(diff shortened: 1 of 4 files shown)


glugovsky (approved):

src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [movies, setMovies] = useState<Movie[]>(moviesFromServer);
+
+  useEffect(() => {
+    const lowerQuery = query.toLowerCase();
+
+    const visibleMovies = moviesFromServer.filter(movie => (
+      movie.title.toLowerCase().includes(lowerQuery)
+        || movie.description.toLowerCase().includes(lowerQuery)
```
OleksiiBlezniuk, line 16: would be nice to move lower casing logic to helper function, but it's not critical

src/App.tsx
```diff
@@ -20,16 +35,20 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={event => (
+                  setQuery(event.target.value)
+                )}
```
OleksiiBlezniuk, line 41: Would be nice to create event handler function instead of setting state here, but in this case it's not critical

OleksiiBlezniuk (approved):

PR #1111, opened 2022-10-28
Develop

frontend112:
https://frontend112.github.io/react_movies-list-filter/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,11 +13,9 @@ const getLowerCase = (expression: string) => (
 );
 
 const handleQueryChange = (query: string) => {
-  const filteredMovies = [...moviesFromServer];
-
   const queryLowerCase = getLowerCase(query);
 
-  return filteredMovies.filter(movie => {
+  return moviesFromServer.filter(movie => {
     return (
       getLowerCase(movie.title)
         .includes(queryLowerCase)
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,45 @@
...
+} from 'react';
+
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-export const App: React.FC = () => {
+interface Movie {
+  title: string;
+  description: string;
```
jstmpelowycz, line 13: I think `description` could be optional

src/App.tsx
```diff
@@ -1,9 +1,45 @@
...
+);
+
+const handleQueryChange = (query: string): Movie[] => {
+  const queryLowerCase = getLowerCase(query);
+
+  return moviesFromServer.filter((movie: Movie) => (
+    getLowerCase(movie.title)
+      .includes(queryLowerCase)
+    || getLowerCase(movie.description)
+      .includes(queryLowerCase)
```
jstmpelowycz, line 30: looks like you duplicate yourself with these pieces of code
```
getLowerCase(...).includes(...)
```

jstmpelowycz: also, if to talk about time complexity (if you don't know about it – google, suggest learning it a little bit more):
```
Array.filter { // O(N)
  String.toLocaleLowerCase // O(N)
  String.includes // O(N)
}

// it can be represented this way – O(N) * (O(N) + O(N)) = 2 * O(N)² ≈ O(N)²
```
and that's a time-consuming operation, try to optimize it a bit if you want, it is not obligatory

jstmpelowycz (approved):
GJ 🔥

PR #1119, opened 2022-11-04
movies-filter

zhenija:
[DEMO LINK](https://zhenija.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,19 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = [...moviesFromServer].filter(movie => {
+    const title = movie.title.toLowerCase().includes(query.toLowerCase());
+    const description = movie.description.toLowerCase()
+      .includes(query.toLowerCase());
+
+    return title || description;
+  });
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,19 @@
...
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = [...moviesFromServer].filter(movie => {
+    const title = movie.title.toLowerCase().includes(query.toLowerCase());
```
loralevitska, line 10: `query.toLowerCase()` - DRY. It's repeatable expression. Move to variable and reuse

src/App.tsx
```diff
@@ -20,12 +30,15 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                onChange={(ev) => {
+                  setQuery(ev.target.value);
+                }}
```
loralevitska, line 35: you can optimize it and move function to variable;
So
const handleChangeQuery = (event) => setQuery(event.target.value)
onChange={handleChangeQuery}

loralevitska (changes requested):

src/App.tsx
zhenija: Then i become an error
image

qoa11a: You need to specify type for an event. For example, React.ChangeEvent\<HTMLInputElement\>.

qoa11a (changes requested):

qoa11a (approved):

PR #1125, opened 2022-11-09
Solution

Fispil:
[DEMO LINK](https://Fispil.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,34 +3,63 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-export const App: React.FC = () => {
-  return (
-    <div className="page">
-      <div className="page-content">
-        <div className="box">
-          <div className="field">
-            {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
-            <label htmlFor="search-query" className="label">
-              Search movie
-            </label>
+type State = {
+  query: string
+};
 
-            <div className="control">
-              <input
-                type="text"
-                id="search-query"
-                className="input"
-                placeholder="Type search word"
-              />
+export class App extends React.Component<{}, State> {
+  state = {
+    query: '',
+  };
+
+  render() {
+    const visibleMovies = moviesFromServer
+      .filter(movie => {
+        const movieTitleCheck = movie.title.toLowerCase()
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -3,34 +3,63 @@ import './App.scss';
...
+                  type="text"
+                  id="search-query"
+                  className="input"
+                  placeholder="Type search word"
+                  value={this.state.query}
+                  onChange={(event) => {
+                    this.setState({
+                      query: event.target.value,
+                    });
+                  }}
```
loralevitska, line 50: move this handler to variable

src/App.tsx
```diff
@@ -3,34 +3,63 @@ import './App.scss';
...
+  render() {
+    const visibleMovies = moviesFromServer
+      .filter(movie => {
+        const movieTitleCheck = movie.title.toLowerCase()
+          .includes(this.state.query.toLowerCase());
+        const movieDescriptionCheck = movie.description.toLowerCase()
+          .includes(this.state.query.toLowerCase());
+
+        return movieTitleCheck || movieDescriptionCheck;
+      });
```
loralevitska, line 24: try to avoid such nesting. You can create auxiliary function in which you will accept the text of the movie and compare it with the entered substring. 
and then you code may look `const visibleMovies = moviesFromServer.filter(movie => compareString(movie))`, 
and take this calculation out of the render method to 14 line

image

loralevitska (changes requested):
need some improvements

src/App.tsx
```diff
@@ -1,16 +1,37 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const movieCheck = (movie: Movie) => {
+    const movieTitleCheck = movie.title.toLowerCase()
+      .includes(query.toLowerCase());
+    const movieDescriptionCheck = movie.description.toLowerCase()
+      .includes(query.toLowerCase());
```
Anastasiia-Svintsova, line 13: Look, you use the same logic twice. Please, create a separate method for it

Anastasiia-Svintsova (changes requested):
One more little fix left

polosanya (approved):
Nice!

PR #1134, opened 2022-11-15
React movie list filter

MeshackDev:
[DEMO LINK](https://MeshackDev.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,32 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,33 @@
...
+    return titleExists || descriptionExists;
+  });
+
+  return visibleMovies;
+}
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = filterMovies(moviesFromServer, query);
```
ihor-jpeg, line 29: this one will be recalculated on every single render
useMemo is the way to go

ihor-jpeg (approved):

PR #1142, opened 2022-11-22
solution

MelnychukYana:
https://MelnychukYana.github.io/react_movies-list-filter/


src/App.tsx
```diff
@@ -20,12 +33,14 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={(event) => handleChange(event.target.value)}
```
stanislav-topikha, line 37:
```suggestion
                onChange={handleChange}
```
make it work like this

stanislav-topikha (changes requested):

PR #1148, opened 2022-11-25
Movie list filter completed

Kry-Gin:
[DEMO LINK](https://kry-gin.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,30 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
+import { Movie } from './types/Movie';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+const filterFilms = (element: string, query: string): boolean => {
+  return element.toLowerCase().includes(query.toLowerCase());
+};
+
+const getFilteredFilms = (
+  films: Movie[],
+  query: string,
+) => {
+  return films.filter(film => {
+    const filterByTitle = filterFilms(film.title, query);
+    const filterByDescription = filterFilms(film.description, query);
+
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,30 @@
...
+  films: Movie[],
+  query: string,
+) => {
+  return films.filter(film => {
+    const filterByTitle = filterFilms(film.title, query);
+    const filterByDescription = filterFilms(film.description, query);
+
+    return filterByTitle || filterByDescription;
+  });
+};
```
referee1988, line 21: Try this syntax. You have 2 extra variables.
```suggestion
const getFilteredFilms = (
  films: Movie[],
  query: string,
) => films.filter(film => (
  filterFilms(film.title, query) || filterFilms(film.description, query)
));
```

src/App.tsx
```diff
@@ -1,9 +1,30 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
+import { Movie } from './types/Movie';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+const filterFilms = (element: string, query: string): boolean => {
```
referee1988, line 7: I believe that the filterFilms() and getFilteredFilms() functions should be written inside a const App component: React.FC.
Both of these functions are inextricably linked to the component and will not be used without it.
It is interesting to know the opinion of the mentor.

referee1988 (approved):
Good job!
There is a small note on extra variables.

src/App.tsx
mykola-koval: agree

mykola-koval (approved):
move the functions inside the app

PR #1160, opened 2022-11-28
movies list filter

Svitlana-Yudina:
[DEMO LINK](https://Svitlana-Yudina.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,28 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
+
+export const findVisibleMovies = (movies: Movie[], query: string): Movie[] => {
+  return movies.filter(movie => {
+    const { description, title } = movie;
+
+    return description.toLocaleLowerCase()
+      .includes(query.toLocaleLowerCase())
+    || title.toLocaleLowerCase()
+      .includes(query.toLocaleLowerCase());
+  });
+};
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = findVisibleMovies(moviesFromServer, query);
+
+  const hendleChangeQuery = ((event: React.ChangeEvent<HTMLInputElement>) => (
+    setQuery(event.target.value)));
+
   return (
     <div className="page">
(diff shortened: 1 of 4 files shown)


viktoriia-ovcharenko (approved):
nice job!

src/App.tsx
```diff
@@ -1,9 +1,28 @@
...
+  });
+};
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = findVisibleMovies(moviesFromServer, query);
+
+  const hendleChangeQuery = ((event: React.ChangeEvent<HTMLInputElement>) => (
+    setQuery(event.target.value)));
```
mykola-koval, line 24: i don't think you need this function

src/App.tsx
```diff
@@ -1,9 +1,28 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
+
+export const findVisibleMovies = (movies: Movie[], query: string): Movie[] => {
+  return movies.filter(movie => {
+    const { description, title } = movie;
+
+    return description.toLocaleLowerCase()
```
mykola-koval, line 11: maybe let's trim query to get better search results?

src/App.tsx
```diff
@@ -1,9 +1,28 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
+
+export const findVisibleMovies = (movies: Movie[], query: string): Movie[] => {
```
mykola-koval, line 7: you can move this function inside the component itself

mykola-koval (changes requested):

src/App.tsx
Svitlana-Yudina: when I write `query.trim().toLocaleLowerCase()` - I have 1 test falied - **should update search results on type**

mykola-koval: yeah, guess we have broken tests
that's okay we can fix this

src/App.tsx
mykola-koval: > should leave it in the "onChange" and not put it in a variable
this ⬆️

mykola-koval (approved):

PR #1164, opened 2022-11-28
movies_list_filter added

Yurii-Pozho:
[DEMO LINK](https://Yurii-Pozho.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,16 @@
-import React from 'react';
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(
+    (movie) => movie.title.toLowerCase().includes(query.toLowerCase())
+      || movie.description.toLowerCase().includes(query.toLowerCase()),
+  );
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,16 @@
...
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(
+    (movie) => movie.title.toLowerCase().includes(query.toLowerCase())
+      || movie.description.toLowerCase().includes(query.toLowerCase()),
+  );
```
Oleksandr-Stryzhavchuk, line 12: maybe it would be better to put lowercase conversion in a separate variable

Oleksandr-Stryzhavchuk: and also it would be better to trim query to get better search results. the mentor left me a comment about this

Oleksandr-Stryzhavchuk (approved):
GJ!

src/App.tsx
```diff
@@ -1,9 +1,16 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(
+    (movie) => movie.title.toLowerCase().includes(query.toLowerCase())
+      || movie.description.toLowerCase().includes(query.toLowerCase()),
```
denys-danyliuk, line 11: you could create helper function for that

denys-danyliuk (approved):

PR #1185, opened 2022-11-30
Solution for list filter

karyna-salnykova:
[DEMO LINK](https://karyna-salnykova.github.io/react_movies-list-filter/)


AlexTsimba (approved):
Great job! I didn't find the reason, why one test is failing. Looks like this condition is satisfied

denys-danyliuk (approved):

PR #1194, opened 2022-12-09
Solution

SergIats:
[DEMO LINK](https://SergIats.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,18 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  let query = '';
+  const [visibleMovies, setVisibleMovies] = useState([...moviesFromServer]);
+
+  const filteredMovies = (movieList: Movie[], searchQuery: string) => {
+    return [...movieList].filter(movie => (
+      movie.description.toLowerCase().includes(searchQuery.toLowerCase())
+      || movie.title.toLowerCase().includes(searchQuery.toLowerCase())));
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,18 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  let query = '';
+  const [visibleMovies, setVisibleMovies] = useState([...moviesFromServer]);
+
+  const filteredMovies = (movieList: Movie[], searchQuery: string) => {
+    return [...movieList].filter(movie => (
+      movie.description.toLowerCase().includes(searchQuery.toLowerCase())
+      || movie.title.toLowerCase().includes(searchQuery.toLowerCase())));
```
nazarmatsevych, line 13: create a separate function for it, since you are using same logic twice

src/App.tsx
```diff
@@ -20,12 +29,16 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                onChange={(event) => {
+                  query = event.target.value;
```
nazarmatsevych, line 33:
```suggestion
                value={query}
                onChange={(event) => {
                  query = event.target.value;
```

nazarmatsevych (changes requested):
Good work,
check comments pls

src/App.tsx
SergIats: Hi, Nazar! Thanks for review. Yes, you're right. I've done.

src/App.tsx
SergIats: Oh...sure. Done)

nazarmatsevych (approved):
great job!

PR #1198, opened 2022-12-14
solved

Olek-Lytvynenko:
[DEMO LINK](https://Olek-Lytvynenko.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,24 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, newQuery] = useState('');
+
+  const searchQuery = (someQuery: string) => (
+    someQuery.toLowerCase().includes(query.toLowerCase())
+  );
+
+  const visibleMovies = moviesFromServer.filter(movie => (
+    searchQuery(movie.title || movie.description)
+  ));
+
   return (
     <div className="page">
       <div className="page-content">
         <div className="box">
           <div className="field">
-            {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
             <label htmlFor="search-query" className="label">
               Search movie
             </label>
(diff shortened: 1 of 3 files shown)


SanyaBratashchuk (changes requested):
image
image
image

src/App.tsx
```diff
@@ -1,15 +1,21 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, newQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => (
+    movie.title.toLowerCase().includes(query.toLowerCase())
+     || movie.description.toLowerCase().includes(query.toLowerCase())
```
nazarmatsevych, line 11: move it to separate function since you are using it twice.

nazarmatsevych (changes requested):

src/App.tsx
```diff
@@ -1,15 +1,24 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, newQuery] = useState('');
+
+  const isIncludes = (value: string) => {
+    return value.toLowerCase().includes(query.toLowerCase());
+  };
```
OlEzhA0, line 11:
```suggestion
  const isIncludes = (value: string) => value.toLowerCase().includes(query.toLowerCase());
```

src/App.tsx
```diff
@@ -20,12 +29,16 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={(event) => {
+                  newQuery(event.target.value);
+                }}
```
OlEzhA0, line 35:
```suggestion
                onChange={(event) => newQuery(event.target.value)}
```

src/App.tsx
```diff
@@ -1,15 +1,24 @@
...
 export const App: React.FC = () => {
+  const [query, newQuery] = useState('');
+
+  const isIncludes = (value: string) => {
+    return value.toLowerCase().includes(query.toLowerCase());
+  };
+
+  const visibleMovies = [...moviesFromServer].filter(movie => (
+    isIncludes(movie.title) || isIncludes(movie.description)
+  ));
```
OlEzhA0, line 15: you may not use spread operator when use `.filter` method

OlEzhA0 (changes requested):
GJ

vlad-demchuk (approved):

PR #1202, opened 2022-12-17
finished

metinbicaksiz:
[DEMO LINK](https://metinbicaksiz.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,23 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter((movie) => {
+    if (
+      movie.title.toLowerCase().includes(query.toString()
+        .toLowerCase()) || movie.description.toLowerCase()
+        .includes(query.toString().toLowerCase())
+    ) {
+      return movie;
+    }
+
+    return false;
+  });
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,23 @@
...
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter((movie) => {
+    if (
+      movie.title.toLowerCase().includes(query.toString()
+        .toLowerCase()) || movie.description.toLowerCase()
+        .includes(query.toString().toLowerCase())
+    ) {
```
jstmpelowycz, line 14: let's make this code construction more readable (either create a variable, or use formatting, or a dedicated function – anything making it simpler to read)

src/App.tsx
```diff
@@ -1,9 +1,23 @@
...
+      return movie;
+    }
+
+    return false;
```
jstmpelowycz, line 18: a callback function being passed to `filter` method should return a boolean, what you return is `Movie | boolean`
let's make it more explicit and simpler, you don't even need to use if-statement

src/App.tsx
```diff
@@ -19,13 +33,18 @@ export const App: React.FC = () => {
                 type="text"
                 id="search-query"
                 className="input"
+                defaultValue={query}
+                // value={filter}
```
jstmpelowycz, line 37: remove such things, you don't need them messing up your code

jstmpelowycz (changes requested):

src/App.tsx
jstmpelowycz: regarding that failing test – have you tried putting `query` inside `value` prop? why are you using `defaultValue` prop?

src/App.tsx
```diff
@@ -19,13 +30,17 @@ export const App: React.FC = () => {
                 type="text"
                 id="search-query"
                 className="input"
+                value={query}
+                onChange={(event) => {
```
polosanya, line 34: You can also create a separate function to handle it

polosanya (approved):
Good job!

PR #1203, opened 2022-12-17
Solution / with a small bug report

UmizDemud:
[DEMO LINK](https://UmizDemud.github.io/react_movies-list-filter/)

A small bug in the tests found and fixed.
In directory /cypress/integration/page.spec.js

Line 114:

```js
  it('should update search results on type', () => {
    page.searchField().type('l');
    page.movies().should('have.length', 5);

    page.searchField().type('o'); // Changed from 'lo' to 'o'
    page.movies().should('have.length', 4);

    page.searchField().type('ve'); // Changed from 'love' to 've'
    page.movies().should('have.length', 3);
  });
```

as the search bar already have the letter 'l' in it, writing *'lo'* makes it '*llo*'

search bar isn't cleared withing the same text in consecutive 'type' actions. 

So I made it that it only types 'l', 'o', 've'. Therefore it searches for the scenario correctly.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,18 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const queryLowerCase = query.trim().toLowerCase();
+  const visibleMovies: Movie[] = moviesFromServer.filter(
+    movie => movie.title.toLowerCase().includes(queryLowerCase)
+      || movie.description.toLowerCase().includes(queryLowerCase),
+  );
+
   return (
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,9 +1,18 @@
...
 import moviesFromServer from './api/movies.json';
+import { Movie } from './types/Movie';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const queryLowerCase = query.trim().toLowerCase();
+  const visibleMovies: Movie[] = moviesFromServer.filter(
+    movie => movie.title.toLowerCase().includes(queryLowerCase)
+      || movie.description.toLowerCase().includes(queryLowerCase),
```
meforoma, line 13: de-structuring + concat title and description could make it bit shorter
```suggestion
    { title, description } => (title + description).toLowerCase().includes(queryLowerCase)
```

meforoma (approved):
nice job!
and a great catch on the flaky test 🚀

PR #1208, opened 2022-12-22
Solution

vitaliidox:
https://vitaliidox.github.io/react_movies-list-filter/


src/App.tsx
```diff
@@ -20,12 +34,13 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                onChange={(event) => setQuery(event.target.value.toLowerCase())}
```
TarasHoliuk, line 37: make this input controlled: add `value={query}`

TarasHoliuk, line 37:
```suggestion
                onChange={(event) => setQuery(event.target.value)}
```
save it in the normal case, you anyway do `.toLowerCase()` in `getSortedMovies`

TarasHoliuk (changes requested):

AlisaVlasova (approved):

PR #1214, opened 2022-12-27
react_movies-list-filter

ysrckr:
[DEMO LINK](https://ysrckr.github.io/react_movies-list-filter/)


UmizDemud (approved):
👍 looks correct

naziamoff (changes requested):
looks good to me 🔥 
still it is better to work on the readability a bit and you're good

As a suggestion - you can try to implement search with debounce here if you want

naziamoff (approved):
Great job 🔥

PR #1228, opened 2023-01-05
solution

halynarydasova:
[DEMO LINK](https://halynarydasova.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,24 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+function filterMovie(query: string): Movie[] {
+  return moviesFromServer.filter(movie => {
+    const title = movie.title.toLowerCase();
+    const description = movie.description.toLowerCase();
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,24 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+function filterMovie(query: string): Movie[] {
```
jstmpelowycz, line 6: good
```suggestion
function filterMovies(query: string): Movie[] {
```
better
```suggestion
function getFilterMovies(query: string): Movie[] {
```

src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
+  return moviesFromServer.filter(movie => {
+    const title = movie.title.toLowerCase();
+    const description = movie.description.toLowerCase();
+    const optimizedQUery = query.toLowerCase().trim();
```
jstmpelowycz, line 10:
```suggestion
    const optimizedQuery = query.toLowerCase().trim();
```

jstmpelowycz (approved):

PR #1247, opened 2023-01-09
React movies list filter

OlexandraKorol:
[DEMO LINK](https://OlexandraKorol.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,16 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,16 @@
...
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = (moviesFromServer.filter((movie) => movie.description
+    .toLowerCase().trim().includes(query.toLowerCase().trim())
+      || movie.title.toLowerCase().trim().includes(query.toLowerCase().trim()))
+  );
```
DiDemkiv, line 12: checklist p2
try to move the same logic to helper functions :)

DiDemkiv (changes requested):

src/App.tsx
OlexandraKorol: I don`t move it to helper, but I try to optimisate this one. What are you thinking?

PR #1250, opened 2023-01-09
solution

DimaKolesnyk:
[DEMO LINK](https://dimaKolesnyk.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,24 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [searchedMovie, setSearchedMovie] = useState('');
+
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [searchedMovie, setSearchedMovie] = useState('');
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setSearchedMovie(event.currentTarget.value);
+  };
+
+  function isIncludes(movie: string) {
```
Oleksandr-Filo, line 13: 1. You should use more understandable name for function, it must show what function actually do (what is includes and where);
2. I think it is better to rename parameter (not movie, probably movieData)

Oleksandr-Filo (approved):
Looks good, just a few things.
If you want, you can improve your filter by adding the ability to ignore extra spaces between entered words

PR #1258, opened 2023-01-09
Solution

vrima1221:
[DEMO LINK](https://vrima1221.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,19 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const normalizedTitle = movie.title.toLowerCase();
+    const normalizedDescription = movie.description.toLowerCase();
+    const normalizedQuery = query.trim().toLowerCase();
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,19 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const normalizedTitle = movie.title.toLowerCase();
+    const normalizedDescription = movie.description.toLowerCase();
+    const normalizedQuery = query.trim().toLowerCase();
```
TheOneThatKnowsNothing, line 11: It would be nice to delete unnecessary spaces between words in the search query. If I search 'love actually' (with two spaces) I will not get results. You can use something like this:
const normalizedQuery = query.toLowerCase().split(' ').filter(Boolean).join(' ');

src/App.tsx
```diff
@@ -1,9 +1,19 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
```
TheOneThatKnowsNothing, line 7: On my opinion it's better to call it searchQuery or searchInput. There could be a lot of different queries in large apps, so it's always nice to use more descriptive variable names, to avoid misunderstanding.

PR #1261, opened 2023-01-09
list filter

anastasiia-roiek:
[DEMO LINK](https://anastasiia-roiek.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,21 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const titleCheck = movie.title.toLowerCase();
+    const descriptionCheck = movie.description.toLowerCase();
+
+    const queryCheck = query.trim().toLowerCase();
+
+    return titleCheck.includes(queryCheck)
+      || descriptionCheck.includes(queryCheck);
+  });
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,21 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const titleCheck = movie.title.toLowerCase();
+    const descriptionCheck = movie.description.toLowerCase();
+
+    const queryCheck = query.trim().toLowerCase();
```
TheOneThatKnowsNothing, line 13: It would be nice to delete unnecessary spaces between words in the search query. If I search 'love    actually' (with two spaces) I will not get results. You can use something like this:
`const queryCheck = query.toLowerCase().split(' ').filter(Boolean).join(' ');`

src/App.tsx
```diff
@@ -1,9 +1,21 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
```
TheOneThatKnowsNothing, line 7: Better call it searchQuery. It's better practice because in large apps could be different queries, so better tell exactly.

src/App.tsx
```diff
@@ -1,9 +1,21 @@
...
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const titleCheck = movie.title.toLowerCase();
+    const descriptionCheck = movie.description.toLowerCase();
+
+    const queryCheck = query.trim().toLowerCase();
+
+    return titleCheck.includes(queryCheck)
+      || descriptionCheck.includes(queryCheck);
```
TheOneThatKnowsNothing, line 16: Names of variables are confusing at this point. 'titleCheck.includes(queryCheck)' better call it 'preparedTitle' or 'formattedTitle' in all cases.

src/App.tsx
Ihor-Karpyn: agree

src/App.tsx
Ihor-Karpyn: agree

Ihor-Karpyn (approved):
Good job, but pay attention to reviewer feedback. It very useful

PR #1267, opened 2023-01-10
Add solution

VladimirEdelmann:
# Task
The `App`contains a `MoviesList` and a search field. Implement filtering using a `useState` hook.
- On every change save the input value into the `query`;
- create a `visibleMovies` variable containing filtered movies;
- check if `movie.title` or `movie.description` contains `query`;
- search should be case insensitive (`Inception` can be found by entering `inc` or `Inc` or even `iNC`).

# Implementation
[DEMO LINK](https://vladimiredelmann.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,31 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const movies = moviesFromServer.filter(movie => {
+    const {
+      title,
+      description,
+    } = movie;
+    const lowerQuery = query.toLowerCase().trim();
+
+    if (
+      title.toLowerCase().includes(lowerQuery)
+        || description.toLowerCase().includes(lowerQuery)
+    ) {
+      return true;
+    }
+
+    return false;
+  });
+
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,31 @@
...
+
+    if (
+      title.toLowerCase().includes(lowerQuery)
+        || description.toLowerCase().includes(lowerQuery)
+    ) {
+      return true;
+    }
+
+    return false;
+  });
```
DiDemkiv, line 23: You can simplify that :)
```suggestion
  return title.toLowerCase().includes(lowerQuery) 
    || description.toLowerCase().includes(lowerQuery);
  });
```

DiDemkiv (changes requested):
👍

DiDemkiv (approved):
Good job! 🚀

src/App.tsx
```diff
@@ -1,9 +1,25 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const movies = moviesFromServer.filter(movie => {
```
valerii-derkach, line 8:
```suggestion
  const [query, setQuery] = useState('');
  
  const movies = moviesFromServer.filter(movie => {
```

valerii-derkach (approved):

PR #1276, opened 2023-01-10
filtered movie list

kreisberg:
[DEMO LINK](https://kreisberg.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,18 +4,18 @@ import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
-  const [movieName, setMovieName] = useState('');
+  const [query, setQuery] = useState('');
 
   const handleOnChange = (event:React.ChangeEvent<HTMLInputElement>) => {
     const { value } = event.target;
 
-    setMovieName(value);
+    setQuery(value);
   };
 
-  const visibleMoviesOnList = moviesFromServer.filter(movie => {
+  const visibleMovies = moviesFromServer.filter(movie => {
     const title = movie.title.toLowerCase();
     const description = movie.description.toLowerCase();
-    const preparedName = movieName.toLowerCase().trim();
+    const preparedName = query.toLowerCase().trim();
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,9 +1,25 @@
...
+  const handleOnChange = (event:React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+
+    setMovieName(value);
+  };
+
+  const visibleMoviesOnList = moviesFromServer.filter(movie => {
+    const title = movie.title.toLowerCase();
+    const description = movie.description.toLowerCase();
+    const name = movieName.toLowerCase().trim();
```
Sveta-Kryukova, line 18: You'd better rename this variable to something like 'preparedName' for better understanding what exactly name you mean)

Sveta-Kryukova (changes requested):

src/App.tsx
kreisberg: Thanks! Just fix it)

src/App.tsx
```diff
@@ -1,9 +1,25 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [movieName, setMovieName] = useState('');
```
valerii-derkach, line 7: we check not only in movieName, but also in the description
```suggestion
  const [query, setQuery] = useState('');
```

src/App.tsx
```diff
@@ -1,9 +1,25 @@
...
+
+  const handleOnChange = (event:React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+
+    setMovieName(value);
+  };
+
+  const visibleMoviesOnList = moviesFromServer.filter(movie => {
```
valerii-derkach, line 15:
```suggestion
  const visibleMovies = moviesFromServer.filter(movie => {
```

valerii-derkach (changes requested):

valerii-derkach (approved):
🔥

PR #1278, opened 2023-01-11
solution

mmahfuzi:
- [DEMO LINK](https://mmahfuzi.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,14 @@
-import React from 'react';
+// import React, { useState } from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setValue] = useState('');
+
+  const filterdMovies = <MoviesList movies={moviesFromServer} value={query} />;
+
   return (
     <div className="page">
       <div className="page-content">
@@ -20,12 +25,17 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={(event) => {
+                  setValue(event.target.value);
+                }}
               />
             </div>
           </div>
-        </div>
+          {query === '' && filterdMovies}
 
-        <MoviesList movies={moviesFromServer} />
+          {query && filterdMovies}
+        </div>
       </div>
 
       <div className="sidebar">
--- a/src/components/MovieCard/MovieCard.tsx
+++ b/src/components/MovieCard/MovieCard.tsx
@@ -3,44 +3,55 @@ import './MovieCard.scss';
 
 interface Props {
   movie: Movie;
+  value: string;
 }
 
-export const MovieCard: React.FC<Props> = ({ movie }) => {
+export const MovieCard: React.FC<Props> = ({ movie, value }) => {
   const {
     imdbUrl, imgUrl, description, title,
   } = movie;
 
-  return (
-    <div className="card">
-      <div className="card-image">
-        <figure className="image is-4by3">
-          <img
-            src={imgUrl}
-            alt="Film logo"
-          />
-        </figure>
-      </div>
-      <div className="card-content">
-        <div className="media">
-          <div className="media-left">
-            <figure className="image is-48x48">
-              <img
-                src="images/imdb-logo.jpeg"
-                alt="imdb"
-              />
-            </figure>
-          </div>
-          <div className="media-content">
-            <p className="title is-8">{title}</p>
-          </div>
-        </div>
+  const valueTrimedLowered = value.trim().toLocaleLowerCase();
 
-        <div className="content">
-          {description}
-          <br />
-          <a href={imdbUrl}>IMDB</a>
+  const check = valueTrimedLowered === title.toLocaleLowerCase()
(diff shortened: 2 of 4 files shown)


src/App.tsx
```diff
@@ -1,9 +1,14 @@
...
+// import React, { useState } from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setValue] = useState('');
+
+  const filterdMovies = <MoviesList movies={moviesFromServer} value={query} />;
```
UmizDemud, line 10: You don't have to define filterdMovies here
You can move the part 
```tsx
<MoviesList movies={moviesFromServer} value={query} />
```

Also check my other comment as it is related

src/App.tsx
```diff
@@ -20,12 +25,17 @@ export const App: React.FC = () => {
...
                 placeholder="Type search word"
+                value={query}
+                onChange={(event) => {
+                  setValue(event.target.value);
+                }}
               />
             </div>
           </div>
-        </div>
+          {query === '' && filterdMovies}
```
UmizDemud, line 35: Here you're checking the query for both 'if its empty, and 'if it's not empty. 
So I thought you can just replace these checks with the part from previous comment

It'd be:

```ts
  </div>
  <MoviesList movies={moviesFromServer} value={query} />
</div>

**Review by UmizDemud — CHANGES_REQUESTED** [review]

File: src/App.tsx (hunk 1 above)
Comment (mmahfuzi) [reply src/App.tsx]: Done

File: src/App.tsx (hunk 2 above)
Comment (mmahfuzi) [reply src/App.tsx]: Done, thanks

**Review by UmizDemud — APPROVED** [review]
Looking great 👍

File: src/components/MovieCard/MovieCard.tsx [hunk 3]
```diff
@@ -3,44 +3,55 @@ import './MovieCard.scss';
...
-              <img
-                src="images/imdb-logo.jpeg"
-                alt="imdb"
-              />
-            </figure>
-          </div>
-          <div className="media-content">
-            <p className="title is-8">{title}</p>
-          </div>
+  const valueTrimedLowered = value.trim().toLocaleLowerCase();
```
naziamoff, line 14: probably better this way?
```suggestion
  const finalQuery = value.trim().toLocaleLowerCase();
```

src/components/MovieCard/MovieCard.tsx
```diff
@@ -3,44 +3,55 @@ import './MovieCard.scss';
...
+
+  const check = valueTrimedLowered === title.toLocaleLowerCase()
```
naziamoff, line 16: check what?

naziamoff: unclear that it is a boolean

src/components/MovieCard/MovieCard.tsx
```diff
@@ -3,44 +3,55 @@ import './MovieCard.scss';
...
+  || title.toLocaleLowerCase().includes(valueTrimedLowered)
+  || description.toLocaleLowerCase().includes(valueTrimedLowered);
+
+  if (check) {
```
naziamoff, line 20: let's render this conditionally, no need to return null

naziamoff (changes requested):
looks good 🔥

src/components/MovieCard/MovieCard.tsx
mmahfuzi: Hello @naziamoff. The ways I tried throw errors, I don't know how to do it :/

src/components/MovieCard/MovieCard.tsx
mmahfuzi: Changed, thanks

src/App.tsx
```diff
@@ -20,12 +22,15 @@ export const App: React.FC = () => {
...
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={(event) => {
+                  setValue(event.target.value);
+                }}
               />
             </div>
           </div>
+          <MoviesList movies={moviesFromServer} value={query} />
```
meforoma, line 32: why not just pass a filtered movies array?
```suggestion
const finalQuery = value.trim().toLocaleLowerCase();
const filteredMovies = moviesFromServer.filter(...
  (title + description).includes(finalQuery)
)
...
          <MoviesList movies={filteredMovies} value={query} />
```

then i guess `MovieCard.tsx` would need no changes at all

meforoma (changes requested):
you've followed an unnecessarily hard path
take a look at suggested filtering approach
i think it's worth refactoring to it

src/App.tsx
mmahfuzi: Hello @meforoma, ok I will apply your suggestion but I don't understand what is the difference? these are two different ways, does the way you are suggesting have some advantages? :) thank you

meforoma: well, yes, the main advantage is
- if you `.map` 5 children components,
- but decide to render them or not inside the child component itself
- it's much less optimal compared to `.map` a list of alreade pre-filtered children components

mmahfuzi: Ohh I see, got it, thanks @meforoma for your explanation :)

meforoma (approved):
✅

PR #1280, opened 2023-01-12
Solution

Grayfox147:
[DEMO LINK](https://Grayfox147.github.io/react_movies-list-filter/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,4 +14,4 @@ The `App`contains a `MoviesList` and a search field. Implement filtering using a
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Use the [React TypeScript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript).
 - Open one more terminal and run tests with `npm test` to ensure your solution is correct.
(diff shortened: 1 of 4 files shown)


mmahfuzi (approved):
Excellent job! Interesting and different approach! I didn't find anything that needs to be changed or corrected. 💯 

Just one thing, I don't get what's the point of Searcbar component, from what I see, I think whatever is inside the Searchbar, can be moved inside the App. Correct me if I'm wrong. :)

Grayfox147:
> Excellent job! Interesting and different approach! I didn't find anything that needs to be changed or corrected. 💯
> 
> Just one thing, I don't get what's the point of Searcbar component, from what I see, I think whatever is inside the Searchbar, can be moved inside the App. Correct me if I'm wrong. :)

yes, basically I just did a functional component for the searchbar to practice, similar like the button Misha showed us before. but theres no need for us to do that or anything still works putting everything directly in the app. :D

naziamoff:
> Excellent job! Interesting and different approach! I didn't find anything that needs to be changed or corrected. 💯
> 
> Just one thing, I don't get what's the point of Searcbar component, from what I see, I think whatever is inside the Searchbar, can be moved inside the App. Correct me if I'm wrong. :)

basically that's the point of the React component approach. Everything should be modal in order to make it easy to reuse

naziamoff (approved):
🔥

PR #1286, opened 2023-01-13
solution

kenvastolord:
[DEMO LINK](https://kenvastolord.github.io/react_movies-list-filter/)


src/App.tsx
```diff
@@ -1,9 +1,19 @@
...
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter((movie) => {
+    const { title, description } = movie;
+    const stringToSearch = `${title} ${description}`;
+
+    return stringToSearch.toLowerCase()
+      .trim()
+      .includes(query.trim().toLowerCase());
+  });
```
dolnaro, line 15: It's probably just me but it looks complicated to use variables in the function expression which is already corresponding to another variable. 
The code below looks easier to read in my opinion.

```
  const visibleMovies = moviesFromServer.filter(movie => (
    ifIncludes(movie.title) || ifIncludes(movie.description)));
```

src/App.tsx
```diff
@@ -20,12 +30,16 @@ export const App: React.FC = () => {
...
               />
             </div>
           </div>
         </div>
 
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={visibleMovies} />
       </div>
 
       <div className="sidebar">
```
dolnaro, line 45: except for that function expression part, everything looks fine :)

dolnaro (approved):

src/App.tsx
meforoma: it could be either way - both are fine

meforoma (approved):
✅

PR #1291, opened 2023-01-13
React movies list (filter)

Mohameddahi:
[DEMO LINK](https://mohameddahi.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,20 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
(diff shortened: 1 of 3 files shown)


SalehBUD (approved):

src/App.tsx
```diff
@@ -1,9 +1,20 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [movies] = useState(moviesFromServer);
```
meforoma, line 8: why would need a state without a state setter function?
i don't think you need one here.
Just `.filter` the `moviesFromServer`
```suggestion
```

meforoma (approved):
Please pay attention to a comment about state without state setter function.

PR #1294, opened 2023-01-13
Develop

SalehAkaJim:
[DEMO LINK](https://SalehBUD.github.io/react_movies-list-filter/)


src/App.tsx
```diff
@@ -1,9 +1,42 @@
...
+  let visibleFilms = [...movies];
+
+  function isIncluded(str: string): boolean {
+    return str
+      .toLocaleLowerCase()
+      .includes(query.trim().toLocaleLowerCase());
+  }
+
+  visibleFilms = visibleFilms.filter((movie) => (
+    isIncluded(movie.title) || isIncluded(movie.description)
```
alaaashkar, line 27: filter method returns a new array, there is no need to create a copy for the Array movies

src/App.tsx
```diff
@@ -17,6 +50,10 @@ export const App: React.FC = () => {
             <div className="control">
               <input
                 type="text"
+                defaultValue={query}
+                onChange={(event) => {
+                  handleSet(event.target.value);
```
alaaashkar, line 55: could be easier to change the value from the setQuery function, the setQuery is already a function that changes the value..

alaaashkar (approved):
just added some comments for more simplicity. 

Great Job !

src/App.tsx
meforoma: totally agree with @alaaashkar - if handler just invokes the setter - just use the setter straight away
```suggestion
                  setQuery(event.target.value);
```

src/App.tsx
```diff
@@ -1,9 +1,42 @@
...
+
+  return visibleFilms;
+}
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  function handleSet(event: string) {
+    setQuery(event);
+  }
```
meforoma, line 38: this handler does seem redundant
```suggestion
```

meforoma (changes requested):
your handler doesn't add any functionality - hence, redundant.
Let's remove it.

meforoma (approved):
✅

PR #1302, opened 2023-01-23
final

Alireza-Mohammadhossein:
[DEMO LINK](https://alireza-mohammadhossein.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,35 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  // const filterMovies = (searchKey: string) => {
+  //   const input = searchKey.toLowerCase().trim();
+
+  //   return moviesFromServer.filter(movie => {
+  //     const title = movie.title.toLowerCase();
+  //     const description = movie.description.toLowerCase();
+
+  //     return title.includes(input) || description.includes(input);
+  //   });
+  // };
+
+  const filterMovies = (searchWord: string) => {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,35 @@
...
+  //     return title.includes(input) || description.includes(input);
+  //   });
+  // };
+
+  const filterMovies = (searchWord: string) => {
+    const word = searchWord.toLowerCase().trim();
+
+    return moviesFromServer.filter(movies => {
+      const title = movies.title.toLowerCase();
+      const description = movies.description.toLowerCase();
```
nurdanokcu, line 25:
```suggestion
    return moviesFromServer.filter(movie => {
      const title = movie.title.toLowerCase();
      const description = movie.description.toLowerCase();
```
because filter method checks every movie one by one in the list
movie would be a more logical name

src/App.tsx
```diff
@@ -1,9 +1,35 @@
...
+  // const filterMovies = (searchKey: string) => {
+  //   const input = searchKey.toLowerCase().trim();
+
+  //   return moviesFromServer.filter(movie => {
+  //     const title = movie.title.toLowerCase();
+  //     const description = movie.description.toLowerCase();
+
+  //     return title.includes(input) || description.includes(input);
+  //   });
+  // };
```
nurdanokcu, line 18: I guess you forgot to remove commented code.

nurdanokcu (changes requested):
Well done,
But I believe code shouldn't be submitted with unnecessary comments.
This is why I will give it another look and approve after the changes.

src/App.tsx
Alireza-Mohammadhossein: I got it Nurdan thanks

nurdanokcu (approved):

naziamoff (approved):
🔥

PR #1303, opened 2023-01-24
Solution

RomeoKs:
[DEMO LINK](https://RomeoKs.github.io/react_movies-list-filter/)


naziamoff (approved):
🔥 
but there are couple of suggestions, take a look

PR #1306, opened 2023-01-25
solution

Malva37:
[DEMO LINK](https://Malva37.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,11 +12,13 @@ export const App: React.FC = () => {
     setQuery(value);
   };
 
+  const preparedQuery = query.trim().toLowerCase();
+
   const visibleMovies = moviesFromServer.filter((movie) => {
     const sutib1 = movie.description.toLowerCase()
-      .includes(query.trim().toLowerCase());
+      .includes(preparedQuery);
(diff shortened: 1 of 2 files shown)


NVyrsta (approved):
Well done (❁´◡`❁)

src/App.tsx
```diff
@@ -1,9 +1,26 @@
...
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+
+    setQuery(value);
+  };
+
+  const visibleMovies = moviesFromServer.filter((movie) => {
+    const sutib1 = movie.description.toLowerCase()
+      .includes(query.trim().toLowerCase());
```
Anna-Tkachenko, line 17: you transform the query several times in your code
let's create separate variable before the filter method with refactored query value, that we'll be able to use here

Anna-Tkachenko (changes requested):

src/App.tsx
Malva37: done)

Anna-Tkachenko (approved):
Good job! ⭐️

PR #1319, opened 2023-01-31
Searchbar implemented

oleh-holovnykh:
[DEMO LINK](https://oleh-holovnykh.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,27 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+// export function
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const { title, description } = movie;
+    const searchQuery = query.trim().toLowerCase();
+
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,27 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+// export function
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const { title, description } = movie;
```
maasiimka, line 11:
```suggestion
  const visibleMovies = moviesFromServer.filter(({ title, description }) => {
```
Good, but you can make destructuring in arguments.

maasiimka (approved):
I don't notice any gross mistakes. You did great job

src/App.tsx
alexandra-protyanova: agree

src/App.tsx
```diff
@@ -1,9 +1,27 @@
...
+    const searchQuery = query.trim().toLowerCase();
+
+    return (title.toLowerCase().includes(searchQuery)
```
alexandra-protyanova, line 14: it would be nice to create separate function for lowerCasing and `includes` check

PR #1324, opened 2023-02-02
react / movies list filter

valntyn:
[DEMO LINK](https://valntyn.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,27 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,27 @@
...
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = e.target;
+
+    setQuery(value);
+  };
+
+  const fixedQuery = query.toLocaleLowerCase().trim();
+
+  const visibleMovies = moviesFromServer.filter(({ title, description }) => {
+    const byName = title.toLocaleLowerCase().includes(fixedQuery);
```
OlEzhA0, line 18: would be better to create something like
```
const isIncludes = (text, query) => 
  text.toLocaleLowerCase().includes(query.toLocaleLowerCase().trim())
```

and usage:
```
return isIncludes(title, query) || isIncludes(description, query)
```

OlEzhA0 (changes requested):
hey! would be better to add few improvements

nazarmatsevych (approved):

PR #1343, opened 2023-02-04
commit message

metalikaqq:
[DEMO LINK](https://metalikaqq.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,24 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [searchParam, setSearchParam] = useState('');
+
+  const movieOnDescktop = moviesFromServer.filter(movie => {
+    const searchParamLowerCase = searchParam.toLocaleLowerCase();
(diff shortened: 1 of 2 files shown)


OlhaYevstifieieva (approved):
Well done 👍🏻

src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
+    const searchParamLowerCase = searchParam.toLocaleLowerCase();
+    const titleLowerCase = movie.title.toLowerCase();
+    const descriptionLowerCase = movie.description.toLowerCase();
+
+    if (titleLowerCase.includes(searchParamLowerCase)
+      || descriptionLowerCase.includes(searchParamLowerCase)) {
+      return movie;
+    }
+
+    return undefined;
```
denys-cheporniuk, line 19: you don't need if here, just return result of includes methods
Also, you can combine titleLowerCase and descriptionLowerCase in one string

src/App.tsx
```diff
@@ -20,12 +35,14 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={searchParam}
+                onChange={event => setSearchParam(event.target.value)}
```
denys-cheporniuk, line 39: move code from onChange to a separate function

denys-cheporniuk (changes requested):

PR #1348, opened 2023-02-05
solution

BrandonSunrise:
[DEMO LINK](https://BrandonSunrise.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,22 @@
-import React from 'react';
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const preparedQuery = query.toLowerCase().trim();
+  const visibleMovies = moviesFromServer.filter(({ title, description }) => {
+    const preparedTitle = title.toLowerCase();
+    const preparedDescription = description.toLowerCase();
+
+    const isIncludeTitle = preparedTitle.includes(preparedQuery);
+    const isIncludeDesription = preparedDescription.includes(preparedQuery);
+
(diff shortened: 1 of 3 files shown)


Coyoote (approved):
Good job!

src/components/MoviesList/MoviesList.tsx
```diff
@@ -6,10 +6,12 @@ interface Props {
...
-);
+export const MoviesList: React.FC<Props> = ({ movies }) => {
+  return (
+    <div className="movies">
+      {movies.map(movie => (
+        <MovieCard key={movie.imdbId} movie={movie} />
+      ))}
+    </div>
+  );
+};
```
alexandra-protyanova, line 17: I think these changes are pointless

src/App.tsx
```diff
@@ -1,9 +1,22 @@
...
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const preparedQuery = query.toLowerCase().trim();
+  const visibleMovies = moviesFromServer.filter(({ title, description }) => {
+    const preparedTitle = title.toLowerCase();
+    const preparedDescription = description.toLowerCase();
+
+    const isIncludeTitle = preparedTitle.includes(preparedQuery);
```
alexandra-protyanova, line 14: it is better to move logic of lowerCasing and `includes` check to separate function

alexandra-protyanova (approved):
Well done 👍

src/components/MoviesList/MoviesList.tsx
BrandonSunrise: Thank you. So it would be better to remove return because it's useless?

src/App.tsx
BrandonSunrise: Roger that!

PR #1351, opened 2023-02-05
movie-list-filter was added

artem-sobolevskyi:
[DEMO LINK](https://artem-sobolevskyi.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -9,10 +9,6 @@ export const App: React.FC = () => {
   const visibleMovies = () => {
     const queryNormilize = query.toLowerCase().trim();
 
-    if (query.length === 0) {
-      return [...moviesFromServer];
-    }
-
     return (moviesFromServer.filter(movie => {
       const movieTitle = movie.title.toLowerCase();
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,33 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
```
alSemenyuk, line 1: I fink you need remove useEffect import

src/App.tsx
```diff
@@ -1,9 +1,33 @@
...
+
+      return (movieTitle.includes(queryNormilize)
+          || movieDescription.includes(queryNormilize)
+      );
+    }));
+  };
+
+  useEffect(() => {
+    filterFilms(query);
+  }, [query]);
```
alSemenyuk, line 29: I don't understand why you use useEffect in this task, because I fink you don't need that

olSemenyuk (changes requested):
GJ, great job but you need make some changes

alexandra-protyanova (approved):
GJ 🚀

PR #1352, opened 2023-02-05
solution

imariash86:
[DEMO LINK](https://imariash86.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,29 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const { title, description } = movie;
+
+    const changeToLowerCase = (word: string) => {
+      return word
+        .split('')
+        .map(letter => letter.toLocaleLowerCase())
+        .join('');
+    };
+
+    const titleToLowerCase = changeToLowerCase(title);
+    const descriptionToLowerCase = changeToLowerCase(description);
+    const queryToLowerCase = changeToLowerCase(query.trim());
+
+    return titleToLowerCase.includes(queryToLowerCase)
+      || descriptionToLowerCase.includes(queryToLowerCase);
+  });
+
   return (
     <div className="page">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,29 @@
...
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const { title, description } = movie;
+
+    const changeToLowerCase = (word: string) => {
```
Oleksandr-Maly, line 12: title, description, and query are strings, so you can use .toLowerCase(), and not to make a special function

Oleksandr-Maly (approved):

src/App.tsx
Oleksandr-Maly: good job!

src/App.tsx
```diff
@@ -1,9 +1,29 @@
...
+      return word
+        .split('')
+        .map(letter => letter.toLocaleLowerCase())
+        .join('');
+    };
```
denys-cheporniuk, line 17: you don't need all these methods, just use toLowerCase for `word`

src/App.tsx
```diff
@@ -20,12 +40,16 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={event => {
+                  setQuery(event.target.value);
+                }}
```
denys-cheporniuk, line 46: move code from onChange to a separate function

src/App.tsx
```diff
@@ -1,9 +1,29 @@
...
+        .map(letter => letter.toLocaleLowerCase())
+        .join('');
+    };
+
+    const titleToLowerCase = changeToLowerCase(title);
+    const descriptionToLowerCase = changeToLowerCase(description);
+    const queryToLowerCase = changeToLowerCase(query.trim());
+
+    return titleToLowerCase.includes(queryToLowerCase)
+      || descriptionToLowerCase.includes(queryToLowerCase);
```
denys-cheporniuk, line 24: you can combine titleToLowerCase and descriptionToLowerCase in one string

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #1366, opened 2023-02-07
solution

miamarichka:


src/App.tsx
```diff
@@ -1,9 +1,24 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const inputHandle = (e: React.ChangeEvent<HTMLInputElement>) => {
```
polosanya, line 9:
```suggestion
  const handleInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
```

src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
+    setQuery(e.currentTarget.value.trim());
+  };
+
+  const normalize = (text:string) => text.toLowerCase();
+
+  const filteredMovies = () => {
+    return [...moviesFromServer].filter(movie => {
+      return normalize(movie.title).includes(query.toLocaleLowerCase())
+        || normalize(movie.description).includes(query.toLocaleLowerCase());
```
polosanya, line 18: Can't you normalize the query using your function?

polosanya (changes requested):
Please don't forget to add a demo link to your PR description, so we can properly check your work

src/App.tsx
miamarichka: I can, just I`ve seen some tutorial, they`ve sad we cant control the text we getting from user, inputs,forms....so I didn`t use it in that way. But now I fixed it)

src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(e.currentTarget.value.trim());
+  };
+
+  const normalize = (text:string) => text.toLowerCase();
```
witflash, line 13: For the future:
this func is just an independent helper so you should move it outside of React Component.

witflash (approved):
Good work.
I left some not critical comments.

PR #1369, opened 2023-02-09
Initial commit

yaroslav-furhala:
- [DEMO LINK](https://yaroslav-furhala.github.io/react_movies-list-filter/)

### DIFF
--- a/src/components/MoviesList/MoviesList.tsx
+++ b/src/components/MoviesList/MoviesList.tsx
@@ -1,15 +1,23 @@
-import React from 'react';
 import './MoviesList.scss';
+import React from 'react';
 import { MovieCard } from '../MovieCard';
 
 interface Props {
   movies: Movie[];
+  query: string;
 }
 
-export const MoviesList: React.FC<Props> = ({ movies }) => (
-  <div className="movies">
-    {movies.map(movie => (
-      <MovieCard key={movie.imdbId} movie={movie} />
-    ))}
-  </div>
-);
+export const MoviesList: React.FC<Props> = ({ movies, query }) => {
+  const lowerQuery = query.toLowerCase();
+  const visibleMovies = movies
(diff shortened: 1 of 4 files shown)


yaroslav-furhala:
Hi, I visited the Q&A and asked a question about this, then I was told that I have a problem with CYPRESS, I tried adding and removing these attributes (visible on the commits) but the result did not change, these last tests are falling for some reason. I do not know what to do with this problem. In fact, the site is working correctly!

src/components/MoviesList/MoviesList.tsx
```diff
@@ -1,15 +1,23 @@
...
-    {movies.map(movie => (
-      <MovieCard key={movie.imdbId} movie={movie} />
-    ))}
-  </div>
-);
+export const MoviesList: React.FC<Props> = ({ movies, query }) => {
+  const lowerQuery = query.toLowerCase();
+  const visibleMovies = movies
+    .filter(movie => movie.title.toLowerCase().includes(lowerQuery)
+    || movie.description.toLowerCase().includes(lowerQuery));
```
Volodymyr-Melnyk-Fron10der, line 14: Try to keep DRY(don't repeat yourself) pattern, I suggest to do smth like that:

```suggestion
  const isQueryMatch = text => text.toLowerCase().includes(query.toLowerCase())

  const visibleMovies = movies
    .filter(({ title, description })=> isQueryMatch(title) || isQueryMatch(description))
```

Volodymyr-Melnyk-Fron10der (changes requested):
Ok, got you👍 Only one improvement in your code then

yaroslav-furhala:
1. (Try to keep DRY(don't repeat yourself) pattern, I suggest to do smth like that:) - Fixed!

SanyaBratashchuk (approved):

PR #1370, opened 2023-02-09
Solution

OksanaBaloh:
[DEMO LINK](https://OksanaBaloh.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,23 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => (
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,23 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => (
+    setQuery(event.target.value)
+  );
+
+  function findMovie(str: string) {
```
ivbutsykin, line 13: Small point. I suggest to use only one code style of functions (arrow, "usual"), you mixed it: `handleChange` declared as arrow function, `findMovie` as "usual". So you can rewrite like:
```
const findMovie = (str: string) => { ... }
```

Also small point is naming. I recommend to rename `findMovie` in `filterMovieCondition`, `filterMovieCallback` or probably you can come up with sth better. Because I expect that I pass this fnc `string` and it return me `movie`, not `bool` value

ivbutsykin (approved):
Good job!

PR #1372, opened 2023-02-09
Solution 1

IVxVI:
[DEMO LINK](https://ivxvi.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,36 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+type Movies = {
+  title: string,
+  description: string,
+  imgUrl: string,
+  imdbUrl: string,
+  imdbId: string,
+};
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  function findMovies(movies: Movies[], input: string) {
+    return movies.filter(
+      ({ title, description }) => {
+        if (input === '') {
+          return movies;
+        }
+
+        return (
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,38 @@
...
+
+        const regex = new RegExp(input.trim(), 'ig');
+
+        return (
+          title.match(regex)
+          || description.match(regex)
+        );
+      },
+    );
+  }
```
Volodymyr-Melnyk-Fron10der, line 32: As I see findMovies hasn`t any dependencies from App component. So you can move it outside App

src/App.tsx
```diff
@@ -1,9 +1,38 @@
...
+  const [query, setQuery] = useState('');
+
+  function findMovies(movies: Movies[], input: string) {
+    return movies.filter(
+      ({ title, description }) => {
+        if (input === '') {
+          return movies;
+        }
+
+        const regex = new RegExp(input.trim(), 'ig');
```
Volodymyr-Melnyk-Fron10der, line 24: If there is no input you shouldn't to do filter loop, but return movies right away. And it's better to trim input to make sure the string is really empty
```suggestion
  function findMovies(movies: Movies[], input: string) {
    if (!input.trim()) {
        return movies;
    }
        
    return movies.filter(({ title, description }) => {
        const regex = new RegExp(input.trim(), 'ig');
```

src/App.tsx
IVxVI: Thanks for the review! I also moved findMovies into separate file, so component looks clearer

src/App.tsx
IVxVI: Thanks, fixed it!

Volodymyr-Melnyk-Fron10der (approved):
Perfect now! I like it👍

PR #1376, opened 2023-02-12
Finish task

U-Ra-All:
[DEMO LINK](https://U-Ra-All.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,22 @@
-import React from 'react';
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,9 +1,22 @@
-import React from 'react';
+import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = [...moviesFromServer].filter(
```
VitaliyBondarenko1982, line 8: For what spreading here?
Filter  is immutable method and return new filtered array.

VitaliyBondarenko1982 (approved):
Good job. Approve.
Just pay attention on one point.

src/App.tsx
U-Ra-All: Yeah, sure, thanks!

PR #1379, opened 2023-02-13
Solution

offsset:
https://offsset.github.io/react_movies-list-filter/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,25 +3,16 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+function setMovie(movieTitle: string) {
+  return moviesFromServer.filter(
+    (movie) => movie.title.toLocaleLowerCase().includes(movieTitle)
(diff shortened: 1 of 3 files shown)


tetiana-ivanchenko (changes requested):
if I search "Love actually" - results is empty, filters should works for all registers

src/App.tsx
```diff
@@ -1,9 +1,28 @@
...
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [visibleMovies, setVisibleMovies] = useState(moviesFromServer);
+
+  function setWords(words: string) {
+    return (words.toLocaleUpperCase().slice(0, 1)
+    + words.toLocaleLowerCase().slice(1));
```
tetiana-ivanchenko, line 12: what is i will search by four words?

tetiana-ivanchenko (changes requested):
when I search smth, and then clear input - I didn't see all films

PR #1391, opened 2023-02-25
finished

barantarasNEW:
https://barantarasNEW.github.io/react_movies-list-filter/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,11 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -19,13 +21,18 @@ export const App: React.FC = () => {
                 type="text"
                 id="search-query"
                 className="input"
+                value={query}
+                onChange={(e) => setQuery(e.target.value)}
```
VitaliyBondarenko1982, line 25: better to make function in component and pass here reference

VitaliyBondarenko1982 (approved):

src/components/MoviesList/MoviesList.tsx
```diff
@@ -4,12 +4,34 @@ import { MovieCard } from '../MovieCard';
...
+      return movies;
+    }
+
+    return movies
+      .filter(({ title, description }) => {
+        return searchWords(title, searchValue)
+          || searchWords(description, searchValue);
+      });
+  };
+
```
etojeDenys, line 27: you can cover it in the useMemo, and use it as a variable by default

etojeDenys (approved):
LGTM

nit:

PR #1395, opened 2023-02-27
React_movies-list-filter

07Spirit:
[DEMO LINK](https://07spirit.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,61 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, inputAskEdit] = useState('');
+
+  const inputInfo: React.ChangeEventHandler<HTMLInputElement> = (event) => {
+    inputAskEdit(event.currentTarget.value);
+  };
+
+  const isFinded = (whereSearch: string[], whatSearch: string[]) => {
+    for (let t = 0; t <= whereSearch.length; t += 1) {
+      if (whereSearch[t] === whatSearch[0]) {
+        for (let i = 0; i <= whatSearch.length; i += 1) {
+          if (i === whatSearch.length) {
+            return true;
+          }
+
+          if (whereSearch[t + i] !== whatSearch[i]) {
+            break;
+          }
+        }
+      } else if (t === whereSearch.length) {
+        return false;
+      }
+    }
+
+    return false;
+  };
+
+  const searchHandler = () => {
+    const filterInputAsk = query.trim().toLowerCase().split('');
+    const isClear = query === '';
+    const visibleMovies = isClear
+      ? [...moviesFromServer]
+      : moviesFromServer.filter((oneOfFilms) => {
+        const titleFormatText = oneOfFilms.title.toLowerCase().split('');
+        const descrFormatText = oneOfFilms.description.toLowerCase().split('');
+
+        return isFinded(
+          titleFormatText,
+          filterInputAsk,
+        ) || isFinded(
+          descrFormatText,
+          filterInputAsk,
+        );
+      });
+
+    return visibleMovies;
+  };
+
   return (
     <div className="page">
       <div className="page-content">
         <div className="box">
           <div className="field">
-            {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,15 +1,61 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, inputAskEdit] = useState('');
```
polosanya, line 7:
```suggestion
  const [query, setQuery] = useState('');
```
set functions should be named just like noun variable, just with "set" prefix

src/App.tsx
```diff
@@ -1,15 +1,61 @@
...
+
+  const inputInfo: React.ChangeEventHandler<HTMLInputElement> = (event) => {
```
polosanya, line 9:
```suggestion
  const handleInputChange: React.ChangeEventHandler<HTMLInputElement> = (event) => {
```

src/App.tsx
```diff
@@ -1,15 +1,61 @@
...
+    inputAskEdit(event.currentTarget.value);
+  };
+
+  const isFinded = (whereSearch: string[], whatSearch: string[]) => {
+    for (let t = 0; t <= whereSearch.length; t += 1) {
+      if (whereSearch[t] === whatSearch[0]) {
+        for (let i = 0; i <= whatSearch.length; i += 1) {
+          if (i === whatSearch.length) {
```
polosanya, line 17: This logic looks too complicated with nested loops.

src/App.tsx
```diff
@@ -1,15 +1,61 @@
...
+  const searchHandler = () => {
+    const filterInputAsk = query.trim().toLowerCase().split('');
+    const isClear = query === '';
+    const visibleMovies = isClear
+      ? [...moviesFromServer]
+      : moviesFromServer.filter((oneOfFilms) => {
+        const titleFormatText = oneOfFilms.title.toLowerCase().split('');
+        const descrFormatText = oneOfFilms.description.toLowerCase().split('');
+
+        return isFinded(
```
polosanya, line 42: Just create a visibleMovies variables that will contain filtered moviesFromServer. You should only check if movie title or movie description includes query, that's all

src/App.tsx
```diff
@@ -20,17 +66,16 @@ export const App: React.FC = () => {
...
                 placeholder="Type search word"
+                value={query}
+                onChange={inputInfo}
               />
             </div>
           </div>
         </div>
-
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={searchHandler()} />
```
polosanya, line 75: You should pass an array of films, not a function there

polosanya (changes requested):
Good job! But for now your solution looks a bit complicated. Let's optimize it. 
Please pay attention to my comments and task description. 
Feel free to ask for help if you have any difficulties)
image

src/App.tsx
07Spirit: Thanks

07Spirit:
My hard drive was crashed, so had a break. :(

src/App.tsx
```diff
@@ -1,15 +1,66 @@
...
+      setVisibleMovies(moviesFromServer);
+    } else {
+      const visibleSingleMovies = moviesFromServer.filter((oneOfFilms) => {
+        const titleFormatText = oneOfFilms.title.toLowerCase().split('');
+        const descrFormatText = oneOfFilms.description
+          .toLowerCase()
+          .split('');
+
+        return (
+          isFinded(titleFormatText, filterInputAsk)
```
witflash, line 44: Why don't you use `includes` method but `for` cycle?
```
const preparedTitle = title.toLowerCase();
const preparedDescription = description.toLowerCase();
const preparedQuery = query.trim().toLowerCase();

return preparedTitle.includes(preparedQuery) || preparedDescription.includes(preparedQuery);
```

witflash (approved):
Well done.
Could you check my comment? Maybe you can simplify your solution.

src/App.tsx
07Spirit: I dont shure about this variant, but my cycle can find overlap. For example, if I write "each", it view form with "each, reach,  eached" etc. But many form find only full word.

07Spirit: Thanks

07Spirit: Your version simply :)

PR #1400, opened 2023-03-03
did the task

vanyachyzh:
[DEMO LINK](https://vanyachyzh.github.io/react_movies-list-filter/)

### DIFF
--- a/src/components/MoviesList/MoviesList.tsx
+++ b/src/components/MoviesList/MoviesList.tsx
@@ -8,11 +8,12 @@ interface Props {
 }
 
 const filtered = (moviesList: Movie[], query: string) => {
-  return moviesList.filter(e => {
-    const lowerTitle = e.title.toLocaleLowerCase();
+  return moviesList.filter(movie => {
+    const lowerDesc = movie.description.toLocaleLowerCase();
+    const lowerTitle = movie.title.toLocaleLowerCase();
     const lowerQuery = query.toLocaleLowerCase();
 
-    return lowerTitle.includes(lowerQuery);
+    return lowerTitle.includes(lowerQuery) || lowerDesc.includes(lowerQuery);
   });
 };
 
--- a/src/components/MoviesList/MoviesList.tsx
+++ b/src/components/MoviesList/MoviesList.tsx
@@ -4,12 +4,28 @@ import { MovieCard } from '../MovieCard';
 
 interface Props {
   movies: Movie[];
+  selectedMovies: string;
 }
 
-export const MoviesList: React.FC<Props> = ({ movies }) => (
-  <div className="movies">
-    {movies.map(movie => (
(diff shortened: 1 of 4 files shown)


src/components/MoviesList/MoviesList.tsx
```diff
@@ -4,12 +4,28 @@ import { MovieCard } from '../MovieCard';
...
+
+  return (
+    <div className="movies">
+      {selectedMovies
+        ? filtered(movies).map(movie => (
+          <MovieCard key={movie.imdbId} movie={movie} />
+        ))
+        : movies.map(movie => (
+          <MovieCard key={movie.imdbId} movie={movie} />
+        ))}
```
IvanFesenko, line 28: Refactor. Create visibleMovies variable and use it for render. If query is not empty it will be equal movies or filtered movies if query is not empty

src/components/Search/Search.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <label htmlFor="search-query" className="label">
+        Search movie
+      </label>
+
+      <div className="control">
+        <input
+          value={selectedMovies}
+          onChange={(e) => {
+            changeMovies(e.target.value);
+          }}
```
IvanFesenko, line 24: create separete handle as was mentioned above

IvanFesenko (changes requested):
Check my comments

src/components/MoviesList/MoviesList.tsx
```diff
@@ -4,12 +4,26 @@ import { MovieCard } from '../MovieCard';
...
 
-export const MoviesList: React.FC<Props> = ({ movies }) => (
-  <div className="movies">
-    {movies.map(movie => (
-      <MovieCard key={movie.imdbId} movie={movie} />
-    ))}
-  </div>
-);
+const filtered = (moviesList: Movie[], query: string) => {
+  return moviesList.filter(e => {
```
VitaliyBondarenko1982, line 11:
```suggestion
  return moviesList.filter(movie=> {
```

VitaliyBondarenko1982: it is not event. better to name movie

src/components/MoviesList/MoviesList.tsx
```diff
@@ -4,12 +4,26 @@ import { MovieCard } from '../MovieCard';
...
+    const lowerTitle = e.title.toLocaleLowerCase();
+    const lowerQuery = query.toLocaleLowerCase();
+
+    return lowerTitle.includes(lowerQuery);
```
VitaliyBondarenko1982, line 15: also should be checking description
Screenshot from 2023-03-03 19-14-43

VitaliyBondarenko1982 (changes requested):

etojeDenys (approved):
LGTM

PR #1403, opened 2023-03-06
solution

NadiiaKotliar:
[DEMO LINK](https://NadiiaKotliar.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,15 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -20,12 +26,15 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                onChange={(event) => {
```
TarasHoliuk, line 29: It's better to move it to a separate function
Just FYI (take it into account in the future)

TarasHoliuk (approved):

PR #1413, opened 2023-03-09
result

novak-k:
[DEMO LINK](https://novak-k.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,37 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+type State = string;
+
+type Movies = {
+  title: string,
+  description: string,
+  imgUrl: string,
+  imdbUrl: string,
+  imdbId: string
+};
+
+export function getFilteredMovies(movies: Movies[], filter: string) {
+  const visibleMovies = [...movies];
+
+  return visibleMovies.filter(movie => movie.title
+    .toLowerCase()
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,37 @@
...
+};
+
+export function getFilteredMovies(movies: Movies[], filter: string) {
+  const visibleMovies = [...movies];
+
+  return visibleMovies.filter(movie => movie.title
+    .toLowerCase()
+    .includes(filter.toLowerCase().trim())
+    || movie.description.toLowerCase().includes(filter.toLowerCase().trim()));
+}
```
Viktor-Kost, line 23: filter is not mutable method
```suggestion
export function getFilteredMovies(movies: Movies[], filter: string) {
  return movies.filter(movie => movie.title
    .toLowerCase()
    .includes(filter.toLowerCase().trim())
    || movie.description.toLowerCase().includes(filter.toLowerCase().trim()));
}
```

src/App.tsx
```diff
@@ -1,9 +1,37 @@
...
+
 export const App: React.FC = () => {
+  const [movieListState, setMovieListState] = useState<State>('');
```
Viktor-Kost, line 26: `movieListState` isn't a meaningful name. It's a filter. So name like "filter", "query" or something

Viktor-Kost (changes requested):
Nice

Need some improvements

src/App.tsx
novak-k: okay

AlisaVlasova (approved):

PR #1415, opened 2023-03-10
Solution

stanislavcodes:
[DEMO LINK](https://stanislavcodes.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,28 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+interface Movie {
+  title: string;
+  description: string;
+  imgUrl: string;
+  imdbUrl: string;
+  imdbId: string;
+}
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies: Movie[] = [...moviesFromServer]
+    .filter(({ title, description }: Movie) => {
+      const keyword = query.toLowerCase().trim();
+      const isInTitle = title.toLowerCase().includes(keyword);
+      const isInDescription = description.toLowerCase().includes(keyword);
+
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -20,12 +39,14 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={(event) => setQuery(event.target.value)}
```
milenadiamonds1, line 43: why don't add helper func and use distructuring like { value } = event.target?

milienadubova (approved):
GJ but let's use helper functions)

src/App.tsx
```diff
@@ -1,9 +1,32 @@
...
+}
+
+export const App: FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies: Movie[] = [...moviesFromServer]
+    .filter(({ title, description }: Movie) => {
+      const keyword = query.toLowerCase().trim();
+      const isInTitle = title.toLowerCase().includes(keyword);
+      const isInDescription = description.toLowerCase().includes(keyword);
```
illiarerun, line 21: you share the same logic  - `toLowerCase` and `includes` between title and description comparison, create helper func for it

src/App.tsx
```diff
@@ -1,9 +1,32 @@
...
+  description: string;
+  imgUrl: string;
+  imdbUrl: string;
+  imdbId: string;
+}
+
+export const App: FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies: Movie[] = [...moviesFromServer]
```
illiarerun, line 17: why do you filter copy?

illiarerun, line 17: also, there's no need for explicitly setting type here
```suggestion
  const visibleMovies = [...moviesFromServer]
```

illiarerun (changes requested):

illiarerun (approved):

PR #1417, opened 2023-03-12
Solution by DK

DimaK123:
[DEMO LINK](https://DimaK123.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,25 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [movies, setMovie] = useState([...moviesFromServer]);
+  const [input, setInput] = useState('');
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+
+    setInput(value);
+
+    const filteredMovies = moviesFromServer.filter(
+      movie => movie.title.toLowerCase().includes(value.toLowerCase().trim())
+      || movie.description.toLowerCase().includes(value.toLowerCase().trim()),
+    );
+
+    setMovie(filteredMovies);
+  };
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,25 @@
...
 export const App: React.FC = () => {
+  const [movies, setMovie] = useState([...moviesFromServer]);
+  const [input, setInput] = useState('');
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+
+    setInput(value);
+
+    const filteredMovies = moviesFromServer.filter(
```
TarasHoliuk, line 15: move it out of `handleChange` function

src/App.tsx
```diff
@@ -17,15 +33,17 @@ export const App: React.FC = () => {
...
                 className="input"
                 placeholder="Type search word"
+                onChange={handleChange}
               />
             </div>
           </div>
         </div>
 
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movies} />
```
TarasHoliuk, line 46:
```suggestion
        <MoviesList movies={filteredMovies} />
```

src/App.tsx
```diff
@@ -1,9 +1,25 @@
...
+    const { value } = event.target;
+
+    setInput(value);
+
+    const filteredMovies = moviesFromServer.filter(
+      movie => movie.title.toLowerCase().includes(value.toLowerCase().trim())
+      || movie.description.toLowerCase().includes(value.toLowerCase().trim()),
+    );
+
+    setMovie(filteredMovies);
```
TarasHoliuk, line 20: you don't need `movies` state - filtered movies will change after every `input` change anyway because it's saved in `state` i.e cause re-render after every change

TarasHoliuk (changes requested):

src/App.tsx
DimaK123: Thanks for advise :)

anstsot (approved):
Great job, approved!

PR #1432, opened 2023-03-20
moviesListFilter1

fenderukr:
https://fenderukr.github.io/react_movies-list-filter


src/App.tsx
```diff
@@ -1,9 +1,30 @@
...
+
+  const getFilteredMovies = () => {
+    if (!query) {
+      return moviesFromServer;
+    }
+
+    const lowerCaseQuery = query.toLowerCase().trim();
+
+    return moviesFromServer.filter(movie => (
+      movie.title.toLowerCase().includes(lowerCaseQuery)
```
Volodymyr-Melnyk-Fron10der, line 17: move this checking into function and apply to both, title and description

Volodymyr-Melnyk-Fron10der (changes requested):
Amazing, just one comment about optimization

src/App.tsx
fenderukr: i had to replace the brach as I was using a new device and had to clone the task again

please check this commit:

9258bfaa7ad1a24f9965714ab0a005f97358d37b

src/App.tsx
```diff
@@ -1,9 +1,32 @@
...
+  const getFilteredMovies = () => {
+    if (!query) {
+      return moviesFromServer;
+    }
+
+    const isMatch
+    = (text: string,
+      textQuery: string) => text.toLowerCase().includes(textQuery);
+
+    const lowerCaseQuery = query.toLowerCase().trim();
```
polosanya, line 18: It would be better to create a function to format the strings instead, as isMatch function doesn't make your code cleaner

src/App.tsx
```diff
@@ -1,9 +1,32 @@
...
+
+    return moviesFromServer.filter(movie => isMatch(movie.title, lowerCaseQuery)
+      || isMatch(movie.description, lowerCaseQuery));
+  };
+
+  const onChangeHandler = (e: ChangeEvent<HTMLInputElement>) => (
```
polosanya, line 24:
```suggestion
  const handleInputChange = (e: ChangeEvent<HTMLInputElement>) => (
```

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments)

PR #1438, opened 2023-03-22
React_movies-list-filter

Anton-Tretiak:
[DEMO LINK](https://anton-tretiak.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+const containQuery = (text: string, query: string) => {
+  return text.toLowerCase().includes(query.toLowerCase());
+};
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const { title, description } = movie;
+
+    return containQuery(title, query.trim())
+      || containQuery(description, query.trim());
+  });
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,22 @@
...
+};
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const { title, description } = movie;
+
+    return containQuery(title, query.trim())
+      || containQuery(description, query.trim());
```
Bogdan-Mykhailov, line 17: you can trim query in your containQuerie function, and not repeat here 2 times this method

src/App.tsx
```diff
@@ -17,15 +30,17 @@ export const App: React.FC = () => {
             <div className="control">
               <input
                 type="text"
+                value={query}
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                onChange={(event) => setQuery(event.target.value)}
```
Bogdan-Mykhailov, line 37: for my opinion better to create handler, and remove all logic from jsx

Bogdan-Mykhailov (approved):
nice

src/App.tsx
```diff
@@ -1,9 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+const containQuery = (text: string, query: string): boolean => {
+  return text.toLowerCase().includes(query.toLowerCase());
```
sergii-nosachenko, line 7:
```suggestion
  return text.toLowerCase().includes(query.trim().toLowerCase());
```

src/App.tsx
sergii-nosachenko, line 17: You can remove the trim from here now.

sergii-nosachenko (changes requested):

src/App.tsx
Anton-Tretiak: I'm not sure, but maybe it's a good point when we have two and more strings. In this situation we can return value(update state) just in one string, and it's not necessarily to create a separate handler.

Bogdan-Mykhailov: You can leave it, but I think it will be much better if there is a minimum of logic in the jsx.

sergii-nosachenko: @Bogdan-Mykhailov Good point

sergii-nosachenko (approved):
Good work 👍

PR #1450, opened 2023-03-23
Develop

VadymTsyndra:
https://VadymTsyndra.github.io/react_movies-list-filter/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,11 +3,7 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-type State = {
-  query: string;
-};
-
-export const App: React.FC<State> = () => {
+export const App: React.FC<{}> = () => {
   const [query, setQuery] = useState('');
 
   const visibleMovies = moviesFromServer.filter(movie => (
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,20 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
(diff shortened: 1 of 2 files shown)


src/components/MoviesList/MoviesList.tsx
```diff
@@ -4,11 +4,14 @@ import { MovieCard } from '../MovieCard';
...
 interface Props {
   movies: Movie[];
+  visibleMovies: Movie[];
 }
 
-export const MoviesList: React.FC<Props> = ({ movies }) => (
+export const MoviesList: React.FC<Props> = ({ movies, visibleMovies }) => (
   <div className="movies">
-    {movies.map(movie => (
+    {visibleMovies ? visibleMovies.map(movie => (
```
Volodymyr-Melnyk-Fron10der, line 12: visibleMovies  is always true, because it's an array , which is truth value. so simplify it

src/App.tsx
```diff
@@ -1,9 +1,16 @@
...
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-export const App: React.FC = () => {
+export const App: React.FC<{}> = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => (
+    movie.title.toLowerCase().includes(query.toLowerCase())
+    || movie.description.toLowerCase().includes(query.toLowerCase())
```
Volodymyr-Melnyk-Fron10der, line 11: create some helper function which is checking does some string passed as an argument include the query

Volodymyr-Melnyk-Fron10der: isInQuey(title) || isInQuery(description)

Volodymyr-Melnyk-Fron10der (changes requested):
Almost done

anstsot (approved):
Great job, approved!

PR #1455, opened 2023-03-24
Solution

arkushenko-anna:
[DEMO LINK](https://arkushenko-anna.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,25 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [movie, setMovie] = useState(moviesFromServer);
+
+  useEffect(() => {
+    const filtred = moviesFromServer.filter(movie => {
+      const { title, description } = movie;
+      const lowerCasedTitle = title.toLowerCase();
+      const lowerCasedDescription  = description.toLowerCase();
+      const lowerCasedQuery = query.toLowerCase().trim();
+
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -20,14 +36,15 @@ export const App: React.FC = () => {
...
                 placeholder="Type search word"
+                value={query}
+                onChange={event => setQuery(event.target.value)}
               />
             </div>
           </div>
         </div>
 
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={movie} />
```
ksavelyeva, line 46: it is better to use naming that give more info than 'movie'

ksavelyeva (approved):

src/App.tsx
```diff
@@ -1,9 +1,25 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [movie, setMovie] = useState(moviesFromServer);
+
+  useEffect(() => {
+    const filtred = moviesFromServer.filter(movie => {
```
Andr1yK, line 11: let's fix linter errors

src/App.tsx
```diff
@@ -1,9 +1,25 @@
...
+      const { title, description } = movie;
+      const lowerCasedTitle = title.toLowerCase();
+      const lowerCasedDescription  = description.toLowerCase();
+      const lowerCasedQuery = query.toLowerCase().trim();
```
Andr1yK, line 15: the same on each iteration, do we need to lowercase it O(n) times?

src/App.tsx
Andr1yK: agree, f.e. `movies`

Andr1yK (changes requested):

Andr1yK (approved):

PR #1477, opened 2023-03-28
React list filter

Sergey-Kovyazin:
[DEMO LINK](https://Sergey-Kovyazin.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,20 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(movie => {
+    return movie.title.toLowerCase().trim().includes(query)
+    || movie.description.toLowerCase().trim().includes(query);
+  });
+
   return (
     <div className="page">
       <div className="page-content">
         <div className="box">
           <div className="field">
-            {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
             <label htmlFor="search-query" className="label">
               Search movie
             </label>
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,15 +1,20 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(movie => {
+    return movie.title.toLowerCase().trim().includes(query)
+    || movie.description.toLowerCase().trim().includes(query);
+  });
```
IvanFesenko, line 11: No need to filter in case, when query is empty

src/App.tsx
```diff
@@ -20,12 +25,14 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={(event) => setQuery(event.target.value)}
```
IvanFesenko, line 29: Create separate handler

IvanFesenko (changes requested):
Lets make some improvements

Sergey-Kovyazin:
One of the test is falling:
 1) Page
       should search by exact title:
     AssertionError: Timed out retrying after 4000ms: Expected to find element: `.card`, but never found it.
      at Context.eval (http://localhost:3001/__cypress/tests?p=cypress/integration/page.spec.js:132:19)
What I need to do to fix it?

src/App.tsx
```diff
@@ -1,15 +1,26 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = query === ''
```
loralevitska, line 8:
```suggestion
  const visibleMovies = !query
```

loralevitska (changes requested):
- If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review
- Fix last point in task 
image

image
- About question try to re-run failed jobs on git (sometimes Git crashes or runs out of memory when building and running tests)

tetiana-ivanchenko (approved):

PR #1480, opened 2023-03-28
Solution

OlhaKissa:
[DEMO LINK](https://olhakissa.github.io/react_movies-list-filter/)

### DIFF
--- a/src/components/MoviesList/MoviesList.tsx
+++ b/src/components/MoviesList/MoviesList.tsx
@@ -1,15 +1,39 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './MoviesList.scss';
 import { MovieCard } from '../MovieCard';
 
 interface Props {
   movies: Movie[];
+  filter: string;
 }
 
-export const MoviesList: React.FC<Props> = ({ movies }) => (
-  <div className="movies">
-    {movies.map(movie => (
-      <MovieCard key={movie.imdbId} movie={movie} />
-    ))}
-  </div>
-);
+export const MoviesList: React.FC<Props> = ({ movies, filter }) => {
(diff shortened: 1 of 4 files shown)


src/components/MoviesList/MoviesList.tsx
```diff
@@ -1,15 +1,39 @@
...
+          || el.description.toLocaleLowerCase()
+            .includes(filter)
+        ) {
+          return el;
+        }
+
+        return null;
+      }));
+    } else {
+      setVisibleMovies([...movies]);
```
mykhalenych, line 28: it is redundant

mykhalenych (changes requested):

src/components/MoviesList/MoviesList.tsx
```diff
@@ -1,15 +1,29 @@
...
 
-export const MoviesList: React.FC<Props> = ({ movies }) => (
-  <div className="movies">
-    {movies.map(movie => (
-      <MovieCard key={movie.imdbId} movie={movie} />
-    ))}
-  </div>
-);
+export const MoviesList: React.FC<Props> = ({ movies, filter }) => {
+  const [visibleMovies, setVisibleMovies] = useState(([...movies]));
```
TarasHoliuk, line 11: no need to create a separate state for values that are dependent on another state value or prop
just create a const variable in this way `const visibleMovies = movies.filter()` - every time movies or filter change component will be re-rendered and the const value will be recalculated

Also, it's better to do it in `App.tsx`, and pass as a prop already filtered `visibleMovies`. The reason: single responsibility principle - `MoviesList` should just render a list and shouldn't know anything about filter

TarasHoliuk (changes requested):

tetiana-ivanchenko (approved):

PR #1481, opened 2023-03-28
done

jaxxxon78:
[DEMO LINK](https://jaxxxon78.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,15 +6,14 @@ import moviesFromServer from './api/movies.json';
 export const App: React.FC = () => {
   const [query, setQuery] = useState('');
 
-  const visibleFilms = moviesFromServer.filter(film => {
-    const description = film.description.toLowerCase();
-    const title = film.title.toLowerCase();
-    const queryChecked = query.toLowerCase().trim();
+  function handler(event: React.ChangeEvent<HTMLInputElement>) {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -19,13 +31,15 @@ export const App: React.FC = () => {
                 type="text"
                 id="search-query"
                 className="input"
+                value={query}
+                onChange={(event) => setQuery(event.target.value)}
```
mykhalenych, line 35: don't use anonymous function

mykhalenych: lowerCase and trim value when you set it

src/App.tsx
```diff
@@ -1,9 +1,21 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleFilms = moviesFromServer.filter(film => {
```
mykhalenych, line 9:
```suggestion
  moviesFromServer.filter(({description, title}) => description.toLowerCase().includes(query) || title.toLowerCase().includes(query));
```

mykhalenych (changes requested):

etojeDenys (approved):
lgtm

PR #1482, opened 2023-03-28
Movies list - Filter solved

sergey-vorobei:
[DEMO LINK](https://sergey-vorobei.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,32 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -20,12 +43,16 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={(event) => {
+                  setQuery(event.target.value);
+                }}
```
TarasHoliuk, line 49: it's better to create a separate function instead of using an anonymous function in such cases - just FYI

TarasHoliuk (approved):

PR #1497, opened 2023-04-08
Solution

DaveBeetle:
[DEMO LINK](https://DaveBeetle.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = [...moviesFromServer].filter((movie) => {
```
ivanchuk-yurii, line 9:
```suggestion
  const visibleMovies = moviesFromServer.filter((movie) => {
```
`filter` doesn't mutate array

ivanchuk-yurii (changes requested):

vlad-demchuk (approved):

PR #1503, opened 2023-04-11
react_movies-list-filter

MaksymHuriev:
[DEMO LINK](https://MaksymHuriev.github.io/react_movies-list-filter/)


loralevitska (changes requested):
Please deploy your solution, and before review check demo_link. If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review
image

PR #1508, opened 2023-04-14
created a movie filter

prokivskaaa:
[DEMO LINK](https://prokivskaaa.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,25 @@
-import React from 'react';
+import React, { ChangeEvent, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+  };
+
+  const visibleMovies = moviesFromServer.filter(({ title, description }) => {
+    const titleLower = title.toLowerCase();
+    const descriptionLower = description.toLowerCase();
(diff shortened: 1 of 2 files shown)


Esteban4769 (approved):
Well done

src/App.tsx
```diff
@@ -1,9 +1,25 @@
...
+  };
+
+  const visibleMovies = moviesFromServer.filter(({ title, description }) => {
+    const titleLower = title.toLowerCase();
+    const descriptionLower = description.toLowerCase();
+    const queryLower = query.trim().toLowerCase();
+
+    return (
+      titleLower.includes(queryLower) || descriptionLower.includes(queryLower)
+    );
```
jstmpelowycz, line 20: DRY, create a function to check if some movie property matches query

src/App.tsx
```diff
@@ -1,9 +1,25 @@
...
+  const [query, setQuery] = useState('');
+
+  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+  };
+
+  const visibleMovies = moviesFromServer.filter(({ title, description }) => {
+    const titleLower = title.toLowerCase();
+    const descriptionLower = description.toLowerCase();
+    const queryLower = query.trim().toLowerCase();
```
jstmpelowycz, line 16: I'd rather call it this way
```js
const lowercasedTitle = ...;
```

jstmpelowycz (changes requested):

src/App.tsx
prokivskaaa: @jstmpelowycz Thank you! Done)

jstmpelowycz (approved):
🔥

PR #1512, opened 2023-04-16
solution

MSA1991:
[DEMO LINK](https://msa1991.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,19 +16,22 @@ export const App: React.FC = () => {
     return str.toLowerCase();
   };
 
-  const handleSearchFilms = (films: Movie[]) => {
+  const searchText = (text: string): boolean => {
     const lowerQuery = getLowerString(query).trim();
 
+    return text.includes(lowerQuery);
+  };
+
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,35 @@
...
+
+  const handleSearchFilms = (films: Movie[]) => {
+    const lowerQuery = getLowerString(query).trim();
+
+    return films.filter(({ title, description }: Movie) => {
+      const lowerTitle = getLowerString(title);
+      const lowerDescription = getLowerString(description);
+
+      return lowerTitle.includes(lowerQuery)
+        || lowerDescription.includes(lowerQuery);
```
Volodymyr-Melnyk-Fron10der, line 27: you can move more logic in the helper function, so code will looks like :

`return helperFunction(title) || helperFunction(description)`

Volodymyr-Melnyk-Fron10der (changes requested):
just one remark to make your solution perfect

qoa11a (approved):

PR #1514, opened 2023-04-18
movies-list-filter

serbinm:
[DEMO LINK](https://serbinm.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -9,12 +9,12 @@ export const App: React.FC = () => {
   const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
     const { value } = event.target;
 
-    setQuery(value);
+    setQuery(value.toLowerCase().trim());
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,22 @@
...
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+
+    setQuery(value.toLowerCase().trim());
+  };
+
+  const moviesData = [...moviesFromServer].filter(movie => {
```
qoa11a, line 15: There's no need to copy an array since the filter method isn't mutating the original array as it creates and returns a copy of an array.

qoa11a (approved):
Perfect! 🤩

PR #1516, opened 2023-04-18
Solution

andriibilous4:
[DEMO LINK](https://andriibilous4.github.io/react_movies-list-filter/)


andriibilous4:
Cannot pass tests - I added trim to query

src/App.tsx
```diff
@@ -3,7 +3,38 @@ import './App.scss';
...
+type Movie = {
+  title: string
+  description: string
+  imgUrl: string
+  imdbUrl: string
+  imdbId: string
+};
+
 export const App: React.FC = () => {
+  const [visiblemovies, setVisibleMovies] = React.useState(moviesFromServer);
```
vlad-demchuk, line 15:
```suggestion
  const [visibleMovies, setVisibleMovies] = React.useState(moviesFromServer);
```

src/App.tsx
```diff
@@ -3,7 +3,38 @@ import './App.scss';
...
+      }
+
+      return prevMovies.filter((movie: Movie) => movie.title
+        .toLowerCase().includes(query)
+        || movie.description.toLowerCase().includes(query));
+    });
+  }, [query]);
+
+  React.useEffect(() => {
+    setVisibleMovies(moviesFromServer as Movie[]);
```
vlad-demchuk, line 35: no need to use these useEffects

you can just create a variable with needed filtered data by query and use it, since you just pass it to a separate component

vlad-demchuk (changes requested):

src/App.tsx
```diff
@@ -3,7 +3,34 @@ import './App.scss';
...
+
 export const App: React.FC = () => {
+  const [visibleMovies, setVisibleMovies] = React.useState(moviesFromServer);
+  const [query, setQuery] = React.useState('');
+
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(e.target.value.trim().toLowerCase());
+  };
+
+  React.useEffect(() => {
```
witflash, line 22: It's better to have a named import for any react hooks.
```
import { useEffect } from 'react';

...

useEffect(() => {
  ...
```

Because if you will have more complex components, your code will be worse readable (you will use too many `React.` repetitions).

witflash (approved):
Good work!
Take attention to my comment for your future work

PR #1518, opened 2023-04-19
Solution

kate-matviiv:


src/App.tsx
```diff
@@ -1,9 +1,17 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const queryLowerCase = query.toLowerCase().trim();
```
polosanya, line 8: Better create this variable inside filter callback, as you use it only there.
Or you can create a function-helper, that will format the string, as you also need to format the movie title and description

polosanya (changes requested):
Almost done! Let's improve your code a bit)

tetiana-ivanchenko (changes requested):
I can't add space, so I can't find movie by two or more words

nazarmatsevych (approved):

PR #1524, opened 2023-04-22
Develop

nazar-medushevskyi:


src/App.tsx
```diff
@@ -1,9 +1,19 @@
...
 
 export const App: React.FC = () => {
+  const [queryInfo, setQueryInfo] = useState('');
+  const visableMovies = moviesFromServer.filter(movie => {
+    const { description, title } = movie;
+
+    const query = queryInfo.trim().toUpperCase();
+
+    return description.toUpperCase().includes(query)
+      || title.toUpperCase().includes(query);
```
Volodymyr-Melnyk-Fron10der, line 14: move this logic into a function 

```suggestion
    return helperFunction(title) || helperFunction(description);
```

Volodymyr-Melnyk-Fron10der (changes requested):
almost great)

src/App.tsx
```diff
@@ -1,9 +1,23 @@
...
 export const App: React.FC = () => {
+  const [queryInfo, setQueryInfo] = useState(' ');
+
+  const helperFunction = (text: string) => {
+    const query = queryInfo.trim().toUpperCase();
+
+    return text.toUpperCase().includes(query);
+  };
+
+  const visableMovies = moviesFromServer.filter(movie => {
```
SanyaBratashchuk, line 15:
```suggestion
  const visibleMovies = moviesFromServer.filter(movie => {
```

src/App.tsx
```diff
@@ -1,9 +1,23 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [queryInfo, setQueryInfo] = useState(' ');
+
+  const helperFunction = (text: string) => {
+    const query = queryInfo.trim().toUpperCase();
+
+    return text.toUpperCase().includes(query);
+  };
```
SanyaBratashchuk, line 13:
```suggestion
  const compareQuery = (text: string) => {
    const query = queryInfo.trim().toUpperCase();

    return text.toUpperCase().includes(query);
  };
```

SanyaBratashchuk (changes requested):

loralevitska (approved):
Good job!

PR #1531, opened 2023-04-24
WorkedVersion

AlexRog01:
[DEMO LINK](https://AlexRog01.github.io/react_movies-list-filter/)

### DIFF
--- a/src/components/MoviesList/MoviesList.tsx
+++ b/src/components/MoviesList/MoviesList.tsx
@@ -4,12 +4,18 @@ import { MovieCard } from '../MovieCard';
 
 interface Props {
   movies: Movie[];
+  query: string;
 }
 
-export const MoviesList: React.FC<Props> = ({ movies }) => (
+export const MoviesList: React.FC<Props> = ({ movies, query }) => (
   <div className="movies">
-    {movies.map(movie => (
-      <MovieCard key={movie.imdbId} movie={movie} />
-    ))}
+    {movies.map(movie => {
+      const isNeededTitle = movie.title.toLowerCase().includes(query.toLowerCase().trim()); // eslint-disable-line
+      const isNeededDesc = movie.description.toLowerCase().includes(query.toLowerCase().trim()); // eslint-disable-line
+
+      return (isNeededTitle || isNeededDesc) && (
+        <MovieCard key={movie.imdbId} movie={movie} />
+      );
+    })}
   </div>
 );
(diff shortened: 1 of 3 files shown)


src/components/MoviesList/MoviesList.tsx
```diff
@@ -4,12 +4,18 @@ import { MovieCard } from '../MovieCard';
...
 
-export const MoviesList: React.FC<Props> = ({ movies }) => (
+export const MoviesList: React.FC<Props> = ({ movies, query }) => (
   <div className="movies">
-    {movies.map(movie => (
-      <MovieCard key={movie.imdbId} movie={movie} />
-    ))}
+    {movies.map(movie => {
+      const isNeededTitle = movie.title.toLowerCase().includes(query.toLowerCase().trim()); // eslint-disable-line
+      const isNeededDesc = movie.description.toLowerCase().includes(query.toLowerCase().trim()); // eslint-disable-line
```
Olekk17, line 14: You could move this logic to separate function and reuse it, because you call lots of same methods here(watch checklist pt. 2)
image
Also I don't think that this is the case where you can disable eslint just because you want to.

Olekk17 (changes requested):

KosolapovAnatolii (approved):
GOOD JOB!

src/components/MoviesList/MoviesList.tsx
```diff
@@ -4,12 +4,22 @@ import { MovieCard } from '../MovieCard';
 
 interface Props {
   movies: Movie[];
+  query: string;
 }
```
jstmpelowycz, line 8: `MovieCard` should know nothing about query, it's another abstraction level
let's keep matching query flow one level above

src/components/MoviesList/MoviesList.tsx
```diff
@@ -4,12 +4,22 @@ import { MovieCard } from '../MovieCard';
...
 
-export const MoviesList: React.FC<Props> = ({ movies }) => (
+function isNeeded(movieInfo: string, query: string) {
```
jstmpelowycz, line 10: stick to descriptive naming
```suggestion
function doesMoviePropertyMatchQuery(movieProperty: string) {
```

jstmpelowycz (changes requested):

SerhiiKirik (approved):
Well done 🚀

PR #1533, opened 2023-04-24
implement search filter

iHateAntichrist:
[DEMO LINK](https://iHateAntichrist.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,32 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [search, setSearch] = useState('');
+
+  const searchFilter = (movie: Movie) => {
+    const { title, description } = movie;
+
+    return title.toLowerCase().includes(search.toLowerCase())
+      || description.toLowerCase().includes(search.toLowerCase());
+  };
+
+  const filteredMovies = moviesFromServer.filter(searchFilter);
+
+  const handleChange = (event: {
+    target: { value: React.SetStateAction<string>; }; }) => {
+    setSearch(event.target.value);
+  };
+
+  const moviesOnScreen = search ? filteredMovies : moviesFromServer;
+
   return (
     <div className="page">
       <div className="page-content">
         <div className="box">
           <div className="field">
-            {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
             <label htmlFor="search-query" className="label">
               Search movie
             </label>
@@ -20,12 +37,14 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                onChange={handleChange}
+                value={search}
               />
             </div>
           </div>
         </div>
 
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={moviesOnScreen} />
       </div>
 
       <div className="sidebar">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,15 +1,32 @@
...
 export const App: React.FC = () => {
+  const [search, setSearch] = useState('');
+
+  const searchFilter = (movie: Movie) => {
+    const { title, description } = movie;
+
+    return title.toLowerCase().includes(search.toLowerCase())
+      || description.toLowerCase().includes(search.toLowerCase());
+  };
+
```
Kavon-Skela, line 15: You didnt handle cases with additional white spaces around the search query. That is why your tests were failed

Kavon-Skela: Also, you have used toLowerCase method two times for the one variable. I think, it will be better if you create additional variable to reuse it

Kavon-Skela (changes requested):
Nice job! You are so close to the target

src/App.tsx
```diff
@@ -1,15 +1,33 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [search, setSearch] = useState('');
+
+  const searchFilter = (movie: Movie) => {
+    const { title, description } = movie;
+    const searchLowerCase = search.toLowerCase().trim();
+
+    return title.toLowerCase().includes(searchLowerCase)
```
SerhiiKirik, line 13: DRY
make a separate function for this & reuse it
Also, split this with variables

src/App.tsx
```diff
@@ -1,15 +1,33 @@
...
+      || description.toLowerCase().includes(searchLowerCase);
+  };
+
+  const filteredMovies = moviesFromServer.filter(searchFilter);
+
+  const handleChange = (event: {
+    target: { value: React.SetStateAction<string>; }; }) => {
+    setSearch(event.target.value);
```
SerhiiKirik, line 21: it's hard to read
Check eventType into `onChange` on `input`
```suggestion
  const handleChange = (event: eventType }) => {
    setSearch(event.target.value);
```

SerhiiKirik (changes requested):
Great 🔥 

src/App.tsx
```diff
@@ -1,15 +1,29 @@
...
 export const App: React.FC = () => {
+  const [search, setSearch] = useState('');
+
+  const toOneCase = (str: string) => str.toLowerCase();
+
+  const searchFilter = (movie: Movie) => {
+    const { title, description } = movie;
+    const searchLowerCase = toOneCase(search.trim());
+
+    return toOneCase(title).includes(searchLowerCase)
```
SerhiiKirik, line 15: split this with variables like - `const formatedTitle = toUserCase(title)`
make function for check if str includes substr

src/App.tsx
```diff
@@ -1,15 +1,29 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [search, setSearch] = useState('');
+
+  const toOneCase = (str: string) => str.toLowerCase();
```
SerhiiKirik, line 9: Move from the component

SerhiiKirik (changes requested):
Great 🔥 

src/App.tsx
```diff
@@ -1,15 +1,33 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [search, setSearch] = useState('');
+
+  function isIncludes(str: string, substr: string): boolean {
```
SerhiiKirik, line 9: move it `func` from the component to avoid creating this `func` in each render

src/App.tsx
```diff
@@ -1,15 +1,33 @@
...
+    const substrLower = substr.toLowerCase().trim();
+
+    return strLower.includes(substrLower);
+  }
+
+  const searchFilter = (movie: Movie) => {
+    const { title, description } = movie;
+
+    return (isIncludes(title, search)
+    || isIncludes(description, search));
```
SerhiiKirik, line 20:
```suggestion
    return isIncludes(title, search) || isIncludes(description, search);
```

SerhiiKirik (approved):
Great 🔥

PR #1534, opened 2023-04-24
Solution

KosolapovAnatolii:
Tests run and failed after 100% complited. Don't understand why -______-

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,21 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+// import { event } from 'cypress/types/jquery';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(({ title, description }) => {
+    return title.toLowerCase().includes(query.toLowerCase())
+    || description.toLowerCase().includes(query.toLowerCase());
+  });
+
   return (
     <div className="page">
       <div className="page-content">
         <div className="box">
           <div className="field">
-            {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
             <label htmlFor="search-query" className="label">
               Search movie
             </label>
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,15 +1,21 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+// import { event } from 'cypress/types/jquery';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(({ title, description }) => {
+    return title.toLowerCase().includes(query.toLowerCase())
+    || description.toLowerCase().includes(query.toLowerCase());
```
vksazonov, line 11: First of all, you have forgotten to check if the user puts spaces in the input
Second, maybe I think it's better to put the check into a separate function and reuse it when needed

vksazonov (changes requested):

src/App.tsx
```diff
@@ -1,15 +1,20 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(({ title, description }) => {
+    return title.toLowerCase().trim().includes(query.toLowerCase())
```
SerhiiKirik, line 9: split it with variables 
also, create a separate function for check `str` includes `substr`
also, you can move normalized these variables into this function

SerhiiKirik (changes requested):
Well done 🔥 

src/App.tsx
```diff
@@ -1,15 +1,29 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(({ title, description }) => {
+    function isIncludes(str: string, substr: string) {
+      return str.includes(substr);
+    }
+
+    const isCoincidence = isIncludes(
```
SerhiiKirik, line 13: move to normalize these strings to the function `isIncludes`
Also, move this func from the component

SerhiiKirik (changes requested):
Great 🔥 

SerhiiKirik (approved):
Well done 🔥

src/App.tsx
```diff
@@ -1,15 +1,26 @@
...
+}
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(({ title, description }) => {
+    const isCoincidence = isIncludes(title.trim(), query)
+    || isIncludes(description, query);
+
+    return isCoincidence;
+  });
```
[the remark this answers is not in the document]
KosolapovAnatolii: feel myself stupid
-___________-

PR #1535, opened 2023-04-24
Solution

Kavon-Skela:
https://Kavon-Skela.github.io/react_movies-list-filter/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,19 +3,30 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-function validationOfString(givenString: string, searchQuery: string) {
-  const lowerQuery = searchQuery.toLowerCase().trim();
+function doesStringMatchQuery(givenString: string, searchQuery: string) {
+  const normalizedQuery = searchQuery.toLowerCase().trim();
 
-  return givenString.toLowerCase().includes(lowerQuery);
+  return givenString.toLowerCase().includes(normalizedQuery);
 }
 
 export const App: React.FC = () => {
   const [query, setQuery] = useState('');
   const visibleMovies: Movie[] = moviesFromServer
-    .filter(({ title, description }) => {
-      return validationOfString(title, query)
-      || validationOfString(description, query);
+    .filter(({
+      title,
+      description,
+    }) => {
+      return doesStringMatchQuery(title, query)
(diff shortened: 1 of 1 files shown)


Sbavdik (approved):
Looks great!

src/App.tsx
```diff
@@ -1,9 +1,18 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies: Movie[] = moviesFromServer
+    .filter(({ title, description }) => {
+      const lowerQuery = query.toLowerCase().trim();
+
+      return title.toLowerCase().includes(lowerQuery)
+      || description.toLowerCase().includes(lowerQuery);
```
jstmpelowycz, line 13: DRY, create a function to check if some movie property matches query

jstmpelowycz (changes requested):

src/App.tsx
```diff
@@ -1,9 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+function validationOfString(givenString: string, searchQuery: string) {
```
jstmpelowycz, line 6: better something like this
```suggestion
function doesStringMatchQuery(givenString: string, searchQuery: string) {
```

src/App.tsx
```diff
@@ -1,9 +1,22 @@
...
+  const lowerQuery = searchQuery.toLowerCase().trim();
```
jstmpelowycz, line 7:
```suggestion
  const normalizedQuery = searchQuery.toLowerCase().trim();
```

src/App.tsx
```diff
@@ -1,9 +1,22 @@
...
+  return givenString.toLowerCase().includes(lowerQuery);
+}
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies: Movie[] = moviesFromServer
+    .filter(({ title, description }) => {
+      return validationOfString(title, query)
+      || validationOfString(description, query);
+    });
```
jstmpelowycz, line 18:
```js

  const visibleMovies: Movie[] = moviesFromServer.filter(({
    title,
    description,
  }) => (
      ...
  ));
```

jstmpelowycz (approved):

PR #1541, opened 2023-04-24
finish

Rage1502:
[DEMO LINK](https://Rage1502.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,27 @@
-import React from 'react';
+import React, { ChangeEvent, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [searchQuery, setSearchQuery] = useState('');
+
+  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    setSearchQuery(event.target.value);
+  };
+
+  const visibleMovies = moviesFromServer
+    .filter(({ title, description }) => {
+      const toTitleLower = title.toLowerCase();
+      const toDescriptionLower = description.toLowerCase();
+      const toQueryLower = searchQuery.trim().toLowerCase();
+
+      return (
+        toTitleLower.includes(toQueryLower)
+          || toDescriptionLower.includes(toQueryLower)
+      );
+    });
+
   return (
     <div className="page">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,27 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [searchQuery, setSearchQuery] = useState('');
+
+  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    setSearchQuery(event.target.value);
+  };
```
roshirin, line 11: I think the function is extra in this case. Would be simplier to use a callback with setSearchQuery in onChange attribute.

PavloReutskiy (approved):
looks good! 👍

src/App.tsx
SerhiiKirik: why?
But what if you want to give some validation, what you will do?

SerhiiKirik (changes requested):
Well done 🚀 

src/App.tsx
```diff
@@ -1,9 +1,31 @@
...
+
+  const сomplianceСheckFields = (field: string, query: string) => {
+    return field.toLowerCase().includes(query.toLowerCase().trim());
+  };
+
+  const visibleMovies = moviesFromServer
+    .filter(({ title, description }) => {
+      // const toTitleLower = title.toLowerCase();
```
SerhiiKirik, line 19: remove extra comments

src/App.tsx
```diff
@@ -1,9 +1,31 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [searchQuery, setSearchQuery] = useState('');
+
+  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    setSearchQuery(event.target.value);
+  };
+
+  const сomplianceСheckFields = (field: string, query: string) => {
```
SerhiiKirik, line 13: move function from component instead to create this function each render
Also, split this with variables

SerhiiKirik (changes requested):
Good job 🔥 
Let's improve your code a bit 🧑‍💻

SerhiiKirik (approved):
Nice 🔥

PR #1545, opened 2023-04-24
add solution

Vlad-Lischenovskiy:
[DEMO LINK](https://Vlad-Lischenovskiy.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,25 @@
-import React from 'react';
+import React, { ChangeEvent, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+  };
+
+  const visibleMovies = moviesFromServer.filter(({ title, description }) => {
+    const titleLower = title.toLowerCase();
+    const descriptionLower = description.toLowerCase();
+    const queryLower = query.toLowerCase().trim();
+
+    return (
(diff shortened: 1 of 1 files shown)


yuliia-anisimova (changes requested):
LGTM 🔥 
Please, check my comment

src/App.tsx
```diff
@@ -1,9 +1,19 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleChange = (movieProperty: string) => {
```
SerhiiKirik, line 9: Move this from the App, bc you create this function each render
Split your ` movieProperty.toLowerCase().includes(query.toLowerCase().trim());` with variables
```suggestion
  const IsIncludes = (string: string, substring: string) => {
```

SerhiiKirik (changes requested):
Great 🚀 

src/App.tsx
Vlad-Lischenovskiy: I'm sorry, I don't understand how to fix this, as you said (

SerhiiKirik: How to move from the component?
Or how to split your looong `movieProperty.toLowerCase().includes(query.toLowerCase().trim())`  with the variables?
```js
 const isIncludes = (string: string, substring: string) => {
    const normalizedString = normalizedString.toLowerCase();
    const normalizedSubstring = query.toLowerCase().trim();

    return normalizedString.includes(normalizedSubstring);
}
```

SerhiiKirik (approved):
Well done 🚀 
Take a look at the comments 🧑‍💻

PR #1546, opened 2023-04-24
completed movie list filter

VenomMB:
[DEMO LINK](https://VenomMB.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,25 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+function isValidFilm(filmProperty: string, query: string): boolean {
+  return filmProperty.toLowerCase().includes(query.toLowerCase().trim());
+}
+
 export const App: React.FC = () => {
+  const [query, changeQuery] = useState('');
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement> &
+  { target : HTMLButtonElement }) => {
+    const { value } = event.target;
+
+    changeQuery(value);
+  };
+
+  const visibleMovies = moviesFromServer.filter(({ title, description }) => (
+    isValidFilm(title, query) || isValidFilm(description, query)));
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,25 @@
...
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+function isValidFilm(filmProperty: string, query: string): boolean {
+  return filmProperty.toLowerCase().includes(query.toLowerCase().trim());
+}
+
 export const App: React.FC = () => {
+  const [query, changeQuery] = useState('');
+
```
glots-a, line 12: In this case, it is better to use a name that begins with the prefix 'set'. This will allow you to have an understanding of Hook's work. (changeQuery => setQury)

glots-a (approved):
GJ👍👍👍

src/App.tsx
```diff
@@ -1,9 +1,25 @@
...
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement> &
```
maksym-didyk, line 13: why do you use this construction in this task?

src/App.tsx
```diff
@@ -1,9 +1,25 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+function isValidFilm(filmProperty: string, query: string): boolean {
+  return filmProperty.toLowerCase().includes(query.toLowerCase().trim());
+}
+
 export const App: React.FC = () => {
+  const [query, changeQuery] = useState('');
```
maksym-didyk, line 11: change changeQuery to setQuery

src/App.tsx
```diff
@@ -20,12 +36,14 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={handleChange}
```
maksym-didyk, line 40: you should use setQuery in this taks (use hooks, not class elements)

maksym-didyk (changes requested):

src/App.tsx
maksym-didyk: maybe correct :)

src/App.tsx
VenomMB: I don't use class elements, handleChange is only a callback which I moved on from onChange.

src/App.tsx
```diff
@@ -1,9 +1,24 @@
-import React from 'react';
+import { useState, ChangeEvent, FC } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-export const App: React.FC = () => {
+function isValidFilm(filmProperty: string, query: string): boolean {
```
jstmpelowycz, line 6:
```suggestion
function doesMoviePropertyMatchQuery(movieProperty: string, query: string): boolean {
```

jstmpelowycz (approved):

PR #1550, opened 2023-04-24
Solution

DimaDvm:
DEMO LINK - https://DimaDvm.github.io/react_movies-list-filter/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const visiableMovies = (query: string) => {
+    return moviesFromServer.filter(movie => {
+      const description = movie.description.toLocaleLowerCase();
+      const title = movie.title.toLocaleLowerCase();
+      const correctQuery = query.toLocaleLowerCase().trim();
+
+      return description.includes(correctQuery)
+        || title.includes(correctQuery);
+    });
+  };
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const visiableMovies = (query: string) => {
```
jstmpelowycz, line 7:
```suggestion
  const getVisibleMovies = (query: string) => {
```

src/App.tsx
```diff
@@ -1,9 +1,22 @@
...
+    return moviesFromServer.filter(movie => {
+      const description = movie.description.toLocaleLowerCase();
+      const title = movie.title.toLocaleLowerCase();
+      const correctQuery = query.toLocaleLowerCase().trim();
+
+      return description.includes(correctQuery)
+        || title.includes(correctQuery);
+    });
```
jstmpelowycz, line 15: DRY, create a function to check if some movie property matches query

src/App.tsx
```diff
@@ -20,12 +33,14 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={event => setQuery(event.target.value)}
```
jstmpelowycz, line 37: you'd rather create a handler

jstmpelowycz (changes requested):

SerhiiKirik (approved):
Nice 🔥

PR #1564, opened 2023-04-25
reactMoviesList Filter

ymys1:
[DEMO LINK](https://ymys1.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,14 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer
+    .filter(movie => (movie.title.toLowerCase().includes(query.toLowerCase()))
+    || (movie.description.toLowerCase().includes(query.toLowerCase())));
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,14 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer
+    .filter(movie => (movie.title.toLowerCase().includes(query.toLowerCase()))
```
SerhiiKirik, line 9: split this with variables
Also, DRY, move this check if `str` includes `substr` to the separate function
image
Trim your query

SerhiiKirik (changes requested):
Great 🔥 

src/App.tsx
```diff
@@ -1,9 +1,21 @@
...
+
+  const filterByTitle = (movie: Movie) => (
+    movie.title.toLowerCase().includes(query.toLowerCase())
+  );
+  const filterByDescription = (movie: Movie) => (
+    movie.description.toLowerCase().includes(query.toLowerCase())
+  );
+
+  const visibleMovies = moviesFromServer
+    .filter((movie) => filterByTitle(movie) || filterByDescription(movie));
```
SerhiiKirik, line 17: 1. It's now a filter
2. You create this function each time
Would be better move to do like that
also, move `isIncludes` from the component to avoid recreate this function
```suggestion
  const isIncludes = (string: string, substring: string) => {
    const normalizedString = string.toLowerCase();
    const normalizedSubstring = substring.toLowerCase().trim();
    
    return normalizedString.includes(normalizedSubstring);
  };

  const visibleMovies = moviesFromServer.filter(({ title, description }) => (
    return isIncludes(title, query) || isIncludes(description ,query)
  );
```

SerhiiKirik (approved):
Great 🔥 
Take a look at the comments 🧑‍💻

PR #1565, opened 2023-04-25
Develop

kukla1989:
[DEMO LINK](https://kukla1989.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,11 +10,11 @@ export const App: React.FC = () => {
       return event.target.value;
     }
 
-    return setQuery(event.target.value.trim());
+    return setQuery(event.target.value);
   };
 
   const VisibleMovies = moviesFromServer.filter(movie => {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,25 @@
...
+      return event.target.value;
+    }
+
+    return setQuery(event.target.value);
+  };
+
+  const VisibleMovies = moviesFromServer.filter(movie => {
+    const queryLower = query.toLowerCase().trim();
+
+    return movie.title.toLowerCase().includes(queryLower)
```
SerhiiKirik, line 19: DRY
spit it with variables & create a separate function for the check if `str` includes `substr`
Also, you can make formating inside this function

SerhiiKirik (changes requested):
Great 🔥 

SerhiiKirik (approved):
Great 🔥

PR #1571, opened 2023-04-27
solution

AnastassiaH:
[DEMO LINK](https://anastassiah.github.io/react_movies-list-filter/)


src/App.tsx
```diff
@@ -1,9 +1,27 @@
...
+//   imdbUrl: string,
+//   imdbId: string,
+// };
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [moviesToRender, setMoviesToRender] = useState(moviesFromServer);
+
+  useEffect(() => {
+    setMoviesToRender([...moviesFromServer].filter(({ title, description }) => {
```
TarasHoliuk, line 19: It's better to create a separate function for filtering movies and call it inside useEffect.

src/App.tsx
```diff
@@ -1,9 +1,27 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+// type Movie = {
```
TarasHoliuk, line 6: You can remove the commented-out code to keep the codebase clean.

src/App.tsx
```diff
@@ -20,12 +38,14 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
```
TarasHoliuk, line 41: Consider trimming the user input before setting the query state to avoid unnecessary re-renders.

TarasHoliuk (changes requested):

src/App.tsx
AnastassiaH: А хіба цее не заборонить ставити пробіл користувачу в інпуті?

tetiana-ivanchenko (approved):

PR #1572, opened 2023-04-28
solution

TanyaS133:
[DEMO LINK](https://TanyaS133.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,32 @@
-import React from 'react';
+import React, { ChangeEvent, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const isPropertyMatchQuery = (movieProperty: string) => (
+    movieProperty
+      .trim()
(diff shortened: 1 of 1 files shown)


KosolapovAnatolii (approved):
Greate!

src/App.tsx
```diff
@@ -1,22 +1,41 @@
...
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const isPropertyMatchQuery = (movieProperty: string) => (
+    movieProperty
+      .trim()
+      .toLocaleLowerCase()
```
liliia-livandovskaya, line 12: Use 'toLowerCase()' instead of 'toLocaleLowerCase()' for better performance, since we don't need locale-specific comparison.

src/App.tsx
```diff
@@ -1,22 +1,41 @@
...
+      .includes(query.toLocaleLowerCase().trim())
```
liliia-livandovskaya, line 13: Use 'toLowerCase()' instead of 'toLocaleLowerCase()' for better performance, since we don't need locale-specific comparison.

liliia-livandovskaya (approved):

PR #1580, opened 2023-05-08
solution

Oleh-Rybachuk:
https://oleh-rybachuk.github.io/react_movies-list-filter/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,29 @@
-import React from 'react';
+import React, { ChangeEvent, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+const foundQuery = (value: string, query: string) => (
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,15 +1,29 @@
...
 import moviesFromServer from './api/movies.json';
 
+const foundQuery = (value: string, query: string) => (
+  value.toLowerCase().includes(query.toLowerCase().trim())
+);
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const foundMovies = moviesFromServer.filter(movie => (
```
Viktor-Kost, line 13: It's better to filter movies only when the query has changed. Consider using [useMemo](https://uk.legacy.reactjs.org/docs/hooks-reference.html#usememo) to do this.

Viktor-Kost (approved):
Good job!
Take a look for the comment I left

PR #1604, opened 2023-05-17
Solution

SuaBy514:
[DEMO LINK](https://suaby514.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,20 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [visibleMovies, setMovies] = useState(moviesFromServer);
+
+  const filterByQuery = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const query = event.target.value.trim(); // trim due to test 15, "should work if user types extra spaces before or after the query"
+
+    setMovies(moviesFromServer.filter((movie) => (
+      movie.title.toLowerCase().includes(query.toLowerCase())
+      || movie.description.toLowerCase().includes(query.toLowerCase())
+    )));
+  };
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,9 +1,20 @@
...
 export const App: React.FC = () => {
+  const [visibleMovies, setMovies] = useState(moviesFromServer);
+
+  const filterByQuery = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const query = event.target.value.trim(); // trim due to test 15, "should work if user types extra spaces before or after the query"
+
+    setMovies(moviesFromServer.filter((movie) => (
+      movie.title.toLowerCase().includes(query.toLowerCase())
+      || movie.description.toLowerCase().includes(query.toLowerCase())
+    )));
```
kasiakubisiak, line 15: lets destructure movie and you can put it to separate function to no rewrite logic

kasiakubisiak (changes requested):

src/App.tsx
```diff
@@ -1,9 +1,26 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+export function filterMovies(query: string) {
+  return (moviesFromServer.filter((movie) => {
+    const { title, description } = movie;
+
+    return (title.toLowerCase().includes(query.toLowerCase())
+    || description.toLowerCase().includes(query.toLowerCase()));
```
kasiakubisiak, line 11: one more thing, put this into separate function not to rewrite logic

kasiakubisiak (changes requested):

src/App.tsx
SuaBy514: I really do not think I understand, let me try to apply this the way I understand it

src/App.tsx
```diff
@@ -1,9 +1,33 @@
...
+  );
+
+  return (moviesFromServer.filter((movie) => {
+    const { title, description } = movie;
+    const formattedTitle = title.toLowerCase();
+    const formattedDescription = description.toLowerCase();
+
+    return (includesQuery(formattedTitle)
+    || includesQuery(formattedDescription));
+  }));
```
d-wasilewski, line 19: Instead of doing it here use .toLowerCase in the `includesQuery` function, it will be easier to manager later because you don't need to remember and manually lower case it every time, the function should take care of it

```suggestion
  return (moviesFromServer.filter((movie) => {
    const { title, description } = movie;

    return (includesQuery(title)
    || includesQuery(description));
  }));
```

d-wasilewski (changes requested):

src/App.tsx
SuaBy514: It lowercases it so that the search is case insensitive, as per the instruction in readme. By removing lowercasing, it causes the following:
image

d-wasilewski (approved):

PR #1619, opened 2023-05-28
react_movie-list-filter done

EhorSaprykin:
[DEMO LINK](https://ehorsaprykin.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -20,17 +33,21 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                // value={query}
```
agniya-i, line 36: Don't comment `value` property
It's important to match the state, that you changing  in `onChange`

src/App.tsx
```diff
@@ -20,17 +33,21 @@ export const App: React.FC = () => {
...
+                onChange={(event) => {
+                  setQuery(event.target.value);
+
+                  return visibleMovie;
```
agniya-i, line 40: Not needed to return anything from this function

agniya-i (changes requested):

PR #1622, opened 2023-05-29
Add React Movies List Filter

ruslanpashkov:
[DEMO LINK](https://ruslanpashkov.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,7 +10,11 @@ export const App: FC = () => {
     const { title, description } = movie;
 
     const movieInfo = `${title} ${description}`.toLowerCase();
-    const normalizedQuery = query.trim().toLowerCase();
+    const normalizedQuery = query
+      .toLowerCase()
+      .split(' ')
+      .filter(Boolean)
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -20,12 +35,16 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={(e) => {
+                  setQuery(e.target.value);
+                }}
```
33Orange, line 41: Create separate function for this event inside your `App` component.
This is a good practice and makes the code easy to maintain.


Checklist p1. Don't use one-letter variable naming
`[CODE STYLE] - Don't use one-letter variable naming. (It could be confusing for another developer, who reads your code. Also, it would really hard to find this variable in the code editor using search)`
```suggestion
                onChange={handleChange}
```

33Orange (changes requested):
Good job.

Just a one moment

silchyn (approved):

PR #1623, opened 2023-05-29
Solution

oleksandr-plias:
[DEMO LINK](https://oleksandr-plias.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,17 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(
+    movie => movie.title.toLocaleLowerCase()
+      .includes(query.toLowerCase().trim())
+    || movie.description.toLocaleLowerCase()
+      .includes(query.toLowerCase().trim()),
+  );
+
   return (
     <div className="page">
       <div className="page-content">
@@ -19,13 +27,15 @@ export const App: React.FC = () => {
                 type="text"
                 id="search-query"
                 className="input"
+                value={query}
                 placeholder="Type search word"
+                onChange={event => setQuery(event.target.value)}
               />
             </div>
           </div>
         </div>
 
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={visibleMovies} />
       </div>
 
       <div className="sidebar">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,17 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const visibleMovies = moviesFromServer.filter(
+    movie => movie.title.toLocaleLowerCase()
```
TarasHoliuk, line 9: you may create variable and save `query.toLowerCase().trim()` to avoid doing it twice
or even better create a function that accepts a string (title or description in this case) to compare it with the query to avoid duplicating the same code twice for title and for description

src/App.tsx
```diff
@@ -19,13 +27,15 @@ export const App: React.FC = () => {
                 type="text"
                 id="search-query"
                 className="input"
+                value={query}
                 placeholder="Type search word"
+                onChange={event => setQuery(event.target.value)}
```
TarasHoliuk, line 32: it's better to create a separate function `handleChange` instead of passing an anonymous function

TarasHoliuk (changes requested):

src/App.tsx
```diff
@@ -1,9 +1,23 @@
...
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const normalizedQuery = query.toLowerCase().trim();
+  const isMatch = (text: string): boolean => {
+    return text.toLowerCase().includes(normalizedQuery);
+  };
+
+  const visibleMovies = moviesFromServer.filter(
+    movie => isMatch(movie.title) || isMatch(movie.description),
+  );
```
VitaliyBodnar, line 15: don't forget about memoization, so  you need to use hook useMemo() here

src/App.tsx
```diff
@@ -1,9 +1,17 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
```
VitaliyBodnar, line 2: according to the style guide, import styles should be at the bottom of the imports list

VitaliyBodnar (changes requested):
Good Job🔥, let’s make a few changes.

src/App.tsx
oleksandr-plias: Maybe this is one of the next topics of study? I have not yet studied this kind of hook

src/App.tsx
oleksandr-plias: corrected

PR #1627, opened 2023-06-05
done

generatedman:
[DEMO LINK](https://generatedman.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { ChangeEvent, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -20,12 +23,25 @@ export const App: React.FC = () => {
...
+                  setVisibleMovies(moviesFromServer.filter(movie => {
+                    return movie
+                      .title
+                      .toLowerCase()
+                      .includes(filmName.toLowerCase()) || movie
+                      .description
+                      .toLowerCase()
+                      .includes(filmName.toLowerCase());
+                  }));
+                }}
```
silchyn, line 38: move this to separate callback and split code into variables and functions to improve readability

silchyn (changes requested):

silchyn (changes requested):

src/App.tsx
```diff
@@ -1,9 +1,24 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [filmName, setFilmName] = useState('');
```
naziamoff, line 7: I would say this is more like a movieSearchQuery

src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
+
+  const visibleMovies = moviesFromServer.filter((movie) => {
+    const { title, description } = movie;
+
+    const movieInfo = `${title} ${description}`.toLowerCase();
+    const normalizedFilmName = filmName
```
naziamoff, line 13: super low-level logic here, don't think this belongs to the main component logic. I would move this one to a separate `helpers.ts` file

src/App.tsx
```diff
@@ -20,12 +35,16 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={filmName}
+                onChange={(event) => {
+                  setFilmName(event.target.value);
+                }}
```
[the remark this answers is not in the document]
naziamoff: agree with Ivan

naziamoff (approved):
not bad. A couple of suggestions though

PR #1636, opened 2023-06-14
React movies filter solution

annperetiatko:
[DEMO LINK]
https://annperetiatko.github.io/react_movies-list-filter/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,29 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [visibleMovies, setFilteredMovies] = useState(moviesFromServer);
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+
+    setQuery(value);
+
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,29 @@
...
+    const filteredMovies = moviesFromServer.filter(
+      movie => movie.title
+        .toLowerCase()
+        .includes(value.toLowerCase().trim())
+      || movie.description
+        .toLowerCase()
+        .includes(value.toLowerCase().trim()),
+    );
+
+    setFilteredMovies(filteredMovies);
```
TarasHoliuk, line 24: don't do it inside `handleChange` function, just set the correct query here
you may create `const filteredMovies` outside `handleChange` function and it will be recalculated (refiltered) every time `query` changes (on every state change)

src/App.tsx
```diff
@@ -1,9 +1,29 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [visibleMovies, setFilteredMovies] = useState(moviesFromServer);
```
TarasHoliuk, line 8: you don't need it, keep only common `filteredMovies` (or rename it to `visibleMovies`) const

TarasHoliuk (changes requested):

PR #1656, opened 2023-06-18
Develop

maryanmats:
[DEMO LINK](https://maryanmats.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,24 @@
-import React from 'react';
+import React, { ChangeEvent, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const { title, description } = movie;
+
+    const isMatch = title.toLowerCase().includes(query.toLowerCase())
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,9 +1,31 @@
...
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const { title, description } = movie;
+
+    const formattedQuery = query.toLowerCase().trim();
+
+    const isInTitle = title
```
silchyn, line 14: What is in title?

src/App.tsx
```diff
@@ -1,9 +1,31 @@
-import React from 'react';
+import React, { ChangeEvent, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
```
silchyn, line 9: So that `filter` is performed each time a component re-renders?

silchyn (changes requested):

src/App.tsx
maryanmats: and how can i fix it?

silchyn (approved):

PR #1659, opened 2023-06-18
Solution

RoksolanaTylipska:
[DEMO LINK](https://RoksolanaTylipska.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,33 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+interface Film {
+  title: string,
+  description: string,
+  imgUrl: string,
+  imdbUrl: string,
+  imdbId: string
+}
+
 export const App: React.FC = () => {
+  const [films, setFilms] = useState<Film[]>(moviesFromServer);
+  const [query, setQuery] = useState('');
+
+  const findFilm = (value: string) => {
+    setQuery(value);
+
+    const lowercasedValue = value.trim().toLowerCase();
+
+    const visibleMovies: Film[] = moviesFromServer.filter(film => (
+      film.title.toLowerCase().includes(lowercasedValue)
+      || film.description.toLowerCase().includes(lowercasedValue)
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -20,12 +44,14 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                value={query}
+                onChange={(event) => findFilm(event.target.value)}
```
silchyn, line 48: Put it in a separate callback

silchyn (changes requested):
So is it a film or a movie?

src/App.tsx
```diff
@@ -1,9 +1,33 @@
...
+
+    const lowercasedValue = value.trim().toLowerCase();
+
+    const visibleMovies: Film[] = moviesFromServer.filter(film => (
+      film.title.toLowerCase().includes(lowercasedValue)
+      || film.description.toLowerCase().includes(lowercasedValue)
+    ));
+
+    setFilms(visibleMovies);
+  };
```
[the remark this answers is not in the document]
RoksolanaTylipska: @silchyn Hello! **'query'** is already declared in the upper scope, I will be use  **'queryString'**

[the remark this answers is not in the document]
RoksolanaTylipska: Hello! '**query**' is already declared in the upper scope, so i decided change to '**queryString**

src/App.tsx
```diff
@@ -1,9 +1,37 @@
...
+interface Film {
+  title: string,
+  description: string,
+  imgUrl: string,
+  imdbUrl: string,
+  imdbId: string
+}
+
 export const App: React.FC = () => {
+  const [movies, setMovies] = useState<Film[]>(moviesFromServer);
```
anstsot, line 15: it will be better to not use `state` if there is possible to avoid it. `useMemo` hook will be better here

anstsot (changes requested):
The code can be improved a bit

src/App.tsx
RoksolanaTylipska: Done

33Orange (approved):
Well done

PR #1666, opened 2023-06-18
Solution

amashutikov:
[DEMO LINK](https://amashutikov.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,36 +1,27 @@
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-interface Event {
-  target: {
-    value: string;
-  }
-}
-
 export const App: React.FC = () => {
   const [query, setQuery] = useState('');
-  const [visibleMovies, setVisibleMovies] = useState(moviesFromServer);
 
-  const handleQueryChange = (event: Event) => {
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
     const { value } = event.target;
 
     setQuery(value);
   };
 
-  useEffect(() => {
-    setVisibleMovies(moviesFromServer.filter(movie => {
-      const searchQuery = query.trim().toLowerCase();
+  const searchQuery = query.trim().toLowerCase();
 
-      const movieInfo = `
-        ${movie.title}
-        ${movie.description}
-      `;
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const isMovieTitleMatch = movie.title.toLowerCase().includes(searchQuery);
+    const isMovieDescriptionMatch = movie.description
+      .toLowerCase()
+      .includes(searchQuery);
 
-      return movieInfo.toLowerCase().includes(searchQuery);
-    }));
-  }, [query]);
+    return isMovieTitleMatch || isMovieDescriptionMatch;
+  });
 
   return (
     <div className="page">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,28 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleQueryChange = (event: { target: { value: string; }; }) => {
```
silchyn, line 9: that's not readable

src/App.tsx
```diff
@@ -1,9 +1,28 @@
...
+    const { value } = event.target;
+
+    setQuery(value);
+  };
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const searchQuery = query.trim().toLowerCase();
+
+    const checkString = `
```
silchyn, line 18:
```suggestion
    const movieInfo = `
```

src/App.tsx
```diff
@@ -1,9 +1,28 @@
...
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleQueryChange = (event: { target: { value: string; }; }) => {
+    const { value } = event.target;
+
+    setQuery(value);
+  };
+
+  const visibleMovies = moviesFromServer.filter(movie => {
```
silchyn, line 15: So that `filter` is performed each time a component re-renders?

silchyn (changes requested):

src/App.tsx
```diff
@@ -1,9 +1,37 @@
...
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+interface Event {
+  target: {
+    value: string;
+  }
+}
+
```
33Orange, line 11:
```suggestion
```

src/App.tsx
```diff
@@ -1,9 +1,37 @@
...
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [visibleMovies, setVisibleMovies] = useState(moviesFromServer);
+
+  const handleQueryChange = (event: Event) => {
```
33Orange, line 16: Let's use react types fo events.

```suggestion
  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```

src/App.tsx
```diff
@@ -1,9 +1,37 @@
...
 
+interface Event {
+  target: {
+    value: string;
+  }
+}
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [visibleMovies, setVisibleMovies] = useState(moviesFromServer);
```
33Orange, line 14: Do you need this state?

src/App.tsx
```diff
@@ -1,9 +1,37 @@
...
+      const searchQuery = query.trim().toLowerCase();
+
+      const movieInfo = `
+        ${movie.title}
+        ${movie.description}
+      `;
+
+      return movieInfo.toLowerCase().includes(searchQuery);
+    }));
+  }, [query]);
```
33Orange, line 33: Let's filter on `moviesFromServer` and use variable with result in render.
You don't need `useEffect` for this.
Move out `searchQuery` from filter to avoid re-creating every iteration.
Let's check title and description separately inside filter and return result.

```suggestion
    const searchQuery = query.trim().toLowerCase();
    setVisibleMovies(moviesFromServer.filter(movie => {
      
      const isMovieTitleMatch = movie.title.toLowerCase().includes(searchQuery);
      ...
    
      return isMovieTitleMatch || ...
    }));
```

33Orange (changes requested):
Almost done

A few moments to fix

33Orange (approved):
Well done

PR #1667, opened 2023-06-18
Solution

MaksimKad:
[DEMO LINK](https://MaksimKad.github.io/react_movies-list-filter/) and add it to the PR description.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,27 @@
-import React from 'react';
+import React, { useState, ChangeEvent } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer
+    .filter(movie => {
+      const lowerCaseQuery = query.toLowerCase().trim();
+
+      return movie.description
+        .toLowerCase()
+        .includes(lowerCaseQuery)
+          || movie.title
+            .toLowerCase()
+            .includes(lowerCaseQuery);
+    });
+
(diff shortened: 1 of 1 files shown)


33Orange (changes requested):
Almost done

src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
+import React, { useState, ChangeEvent } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const preparedQuery = query.toLowerCase().trim();
```
dffUqp, line 10: Let's move this `preparedQuery` out of the filter method.😃
We need it to avoid creating a new variable at each iteration.

src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
+    const preparedTitle = movie.title.toLowerCase();
+    const preparedDesc = movie.description.toLowerCase();
+
+    return preparedTitle.includes(preparedQuery)
+        || preparedDesc.includes(preparedQuery);
+  });
+
+  const changeSearchInput = (event: ChangeEvent<HTMLInputElement>) => {
```
dffUqp, line 18: good practice to name the event handler as handle[eventName]
```suggestion
  const handleInputChange = (event: ChangeEvent<HTMLInputElement>) => {
```

dffUqp (changes requested):
LGTM but needs a few fixes

dffUqp (approved):
it seems you forgot to push fixes but I'm sure that you fixed all comments so I will approve this

PR #1673, opened 2023-06-19
made react task movies-list-filter

vladyslav-lapshyn:
[DEMO LINK](https://vladyslav-lapshyn.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,28 @@
-import React from 'react';
+import { FC, useEffect, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
-export const App: React.FC = () => {
+export const App: FC = () => {
+  const [visibleMovies, setVisibleMovies] = useState(moviesFromServer);
+  const [query, setQuery] = useState('');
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,9 +1,28 @@
...
+    const filteredMovies = moviesFromServer.filter((movie) => {
+      const lowerCaseSearchQuery = query.toLowerCase().trim();
+
+      return [movie.title, movie.description].some(
+        (text) => text.toLowerCase().includes(lowerCaseSearchQuery),
+      );
+    });
+
+    setVisibleMovies(filteredMovies);
+  }, [query]);
```
dffUqp, line 24: This code look nice but let's simplify it
```suggestion
    const lowerCaseSearchQuery = query.toLowerCase().trim();
    
    const visibleMovies = moviesFromServer.filter(movie => {
      return [movie.title, movie.description].some(
        (text) => text.toLowerCase().includes(lowerCaseSearchQuery),
      );
  })
```

dffUqp (changes requested):
LGTM but needs a few fixes

33Orange (approved):
Well done

PR #1674, opened 2023-06-19
add movies-list filter solution

vchervinko:
[DEMO LINK](https://vchervinko.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,25 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,25 @@
...
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(movie => {
+    const { title, description } = movie;
+    const preparedTitle = title.toLowerCase();
+    const preparedDescription = description.toLowerCase();
+    const preparedQuery = query.trim().toLowerCase();
```
dffUqp, line 13: Let's move this variable out of the filter method.😃
We need it to avoid creating a new variable at each iteration.

dffUqp (changes requested):
LGTM but needs a few fixes

33Orange (approved):
Well done

PR #1678, opened 2023-06-19
added task solution

holytimy4:
[DEMO LINK](https://holytimy4.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,34 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const filteredMovies = moviesFromServer.filter(movie => {
+    const {
+      title,
+      description,
+    } = movie;
+
+    const searchRegex = new RegExp(query.trim(), 'gi');
+    const isMatch = searchRegex.test(title) || searchRegex.test(description);
+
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,15 +1,34 @@
...
+    const {
+      title,
+      description,
+    } = movie;
+
+    const searchRegex = new RegExp(query.trim(), 'gi');
+    const isMatch = searchRegex.test(title) || searchRegex.test(description);
+
+    return isMatch;
+  });
```
33Orange, line 19: Let's move out creating `searchRegex` variable to avoid re-creating every iteration in filter

```suggestion
  const searchRegex = new RegExp(query.trim(), 'gi');
  
  const filteredMovies = moviesFromServer.filter(movie => {
    const {
      title,
      description,
    } = movie;
    
    const isMatch = searchRegex.test(title) || searchRegex.test(description);

    return isMatch;
  });
```

33Orange (changes requested):
Almost done

Just a one moment to fix

src/App.tsx
```diff
@@ -1,15 +1,34 @@
...
+
+  const searchInputHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+
+    setQuery(value);
+  };
```
silchyn, line 25:
```suggestion
  const handleInput = (event: React.ChangeEvent<HTMLInputElement>) => {
    setQuery(event.target.value);
  };
```

silchyn (approved):

PR #1686, opened 2023-06-26
solution

Spec0s:
would like to note that in your working version there is no trim():
https://mate-academy.github.io/react_movies-list-filter/

my working:
https://spec0s.github.io/react_movies-list-filter/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,20 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const matchesSearchQuery = (words: string) => words
+    .toLowerCase()
+    .includes(query.trim().toLowerCase());
+
+  const visibleMovies = moviesFromServer.filter(movie => (
+    matchesSearchQuery(movie.title)
+    || matchesSearchQuery(movie.description)
+  ));
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -20,12 +31,15 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                onChange={(word) => {
+                  setQuery(word.target.value);
+                }}
```
Anastasiia-Svintsova, line 36: It's better to use event instead of word in the onChange event handler because event is a commonly used and recognized parameter name for event handlers in JavaScript.

src/components/MoviesList/MoviesList.tsx
```diff
@@ -8,8 +8,9 @@ interface Props {
 
 export const MoviesList: React.FC<Props> = ({ movies }) => (
   <div className="movies">
-    {movies.map(movie => (
-      <MovieCard key={movie.imdbId} movie={movie} />
-    ))}
+    {movies
+      .map(movie => (
```
Anastasiia-Svintsova, line 12:
```suggestion
    {movies.map(movie => (
```

src/App.tsx
```diff
@@ -1,9 +1,20 @@
...
+  const [query, setQuery] = useState('');
+
+  const matchesSearchQuery = (words: string) => words
+    .toLowerCase()
+    .includes(query.trim().toLowerCase());
+
+  const visibleMovies = moviesFromServer.filter(movie => (
+    matchesSearchQuery(movie.title)
+    || matchesSearchQuery(movie.description)
+  ));
```
Anastasiia-Svintsova, line 16: You can use destructuring for the movie object to improve the code readability

Anastasiia-Svintsova (approved):

PR #1704, opened 2023-07-12
Solution

superzuc:
https://superzuc.github.io/react_movies-list-filter/


src/App.tsx
```diff
@@ -1,9 +1,23 @@
...
+
+  const handleSearch = (event: { target:
+  { value: React.SetStateAction<string>; }; }) => {
+    searchFilm(event.target.value);
+  };
+
+  const [movies] = useState(moviesFromServer);
+
+  const filteredMovies = movies.filter((movie) =>
+    // eslint-disable-next-line implicit-arrow-linebreak
```
polosanya, line 17: Avoid disabling ESLint rules unless it is absolutely necessary. In this case, the line break can be avoided by wrapping the return statement of the arrow function in parentheses.

src/App.tsx
```diff
@@ -1,9 +1,23 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [filmTitle, searchFilm] = useState('');
```
polosanya, line 7:
```suggestion
  const [searchQuery , setSearchQuery] = useState('');
```
The name is a bit confusing as it's the value which the user types in input, not necessarily the film's title

polosanya (changes requested):
Almost done! Try to pass the last test
image

src/App.tsx
```diff
@@ -1,9 +1,22 @@
...
 
 export const App: React.FC = () => {
+  const [searchQuery, setSearchQuery] = useState('');
+
+  const handleSearch = (event: { target:
+  { value: React.SetStateAction<string>; }; }) => {
+    setSearchQuery(event.target.value);
+  };
+
+  const [movies] = useState(moviesFromServer);
```
lerastarynets, line 14: What is the purpose of this state? Why not passing `moviesFromServer` to filteredMovies directly?

lerastarynets (changes requested):
Almost there) Just one more thing needs to be fixed

PR #1706, opened 2023-07-13
solution

mariuszgit:


src/types/Movie/Movie.ts
```diff
@@ -0,0 +1,7 @@
+export interface MovieType {
```
DorotaLeniecDev, line 1: No need to name it **MovieType**, **Movie** is enought

src/App.tsx
```diff
@@ -1,9 +1,21 @@
...
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+import { MovieType } from './types/Movie/Movie';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const filteredMovies = (moviesList: MovieType[], searchQuery: string) => {
```
DorotaLeniecDev, line 10:
```suggestion
  const filterMovies = (moviesList: MovieType[], searchQuery: string) => {
```

DorotaLeniecDev (approved):
2 name suggestion changes but this is looking good, approving 👌

PR #1711, opened 2023-08-15
add task solution

Luk2asz:
[DEMO LINK](https://Luk2asz.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,15 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleChange: React.ChangeEventHandler<HTMLInputElement> = (event) => {
+    setQuery(event.target.value);
+  };
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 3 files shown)


src/components/MoviesList/MoviesList.tsx
darokrk, line 1: you don't need to create this file, it's already created in task description, you should only handle logic for this component.

please remove it bcs it is redundant

src/App.tsx
```diff
@@ -20,12 +26,13 @@ export const App: React.FC = () => {
...
                 className="input"
                 placeholder="Type search word"
+                onChange={handleChange}
               />
             </div>
           </div>
         </div>
 
-        <MoviesList movies={moviesFromServer} />
+        <MoviesList movies={moviesFromServer} query={query} />
```
darokrk, line 35: you should only pass here as props movies as filtered moviesFromServer, query is redundant so please remove it

src/App.tsx
```diff
@@ -1,9 +1,15 @@
...
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleChange: React.ChangeEventHandler<HTMLInputElement> = (event) => {
+    setQuery(event.target.value);
+  };
+
```
darokrk, line 12: you should add logic here, create variable e.g. `const visibleMovies` which will be passed as props `movies={visibleMovies}` to your `MoviesList` component, you should use the filter method to handle this logic based on what you stored in `query` state, 

```
const visibleMovies = moviesFromServer
    .filter(etc..... logic here)
```

darokrk (changes requested):
Mentions left, the logic of component need to be fixed to passed task requirements 🛠️ 👨🏽‍💻

mykhalenych (approved):

PR #1717, opened 2023-09-04
solutionn

Dolnys:
[DEMO LINK](https://Dolnys.github.io/react_movies-list-filter/)


choeqq (approved):
Looks good, please read about [input debouncing](https://dev.to/manishkc104/debounce-input-in-react-3726), it's common practice every frontend app

PR #1721, opened 2023-09-06
Add solution

WaveOfDandelions:
GH: https://WaveOfDandelions.github.io/react_movies-list-filter/


src/App.tsx
```diff
@@ -1,9 +1,26 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
```
choeqq, line 8: I think I would have another state for currently visibleMovies, just to be clear what's is currently displayed

src/App.tsx
```diff
@@ -19,13 +36,17 @@ export const App: React.FC = () => {
                 type="text"
                 id="search-query"
                 className="input"
+                value={query}
                 placeholder="Type search word"
+                onChange={(event) => {
+                  setQuery(() => event.target.value);
+                }}
```
choeqq, line 43:
```suggestion
                onChange={handleInputChange}
```

choeqq (changes requested):

src/App.tsx
WaveOfDandelions: > 

Fixed

src/App.tsx
WaveOfDandelions: > I think I would have another state for currently visibleMovies, just to be clear what's is currently displayed

Please provide example if you can

choeqq (approved):

PR #1741, opened 2023-09-11
Task solution

ArtemBehter:


src/App.tsx
```diff
@@ -1,9 +1,24 @@
...
+    return moviesFromServer.filter(movie => {
+      const title = movie.title.toLowerCase();
+      const description = movie.description.toLowerCase();
+
+      return title.includes(query) || description.includes(query);
+    });
+  };
+
+  const searchedValue = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value.toLocaleLowerCase().trim());
```
darokrk, line 19: The trimming of input value may lead to unexpected results. Consider removing the trim method.

src/App.tsx
```diff
@@ -1,9 +1,24 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const sortedMovies = () => {
```
darokrk, line 9: you can refactor this to store result in variable, e.g.:

```
const sortedMovies = moviesFromServer.filter(movie => {
      const title = movie.title.toLowerCase();
      const description = movie.description.toLowerCase();

      return title.includes(query) || description.includes(query);
    });
```

then you avoid using function call is MoviesList props so this increases performance and avoid unexpected re-renders calls:

`<MoviesList movies={sortedMovies} />`

darokrk (changes requested):
Looks good, few mentions left for code improvement 🛠️

darokrk (approved):
Good job 👏🏽  🏆

PR #1745, opened 2023-09-27
done

Alessandranitti15:
[DEMO LINK](https://Alessandranitti15.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,26 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,26 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const movies = moviesFromServer;
```
lerastarynets, line 7: why not just use `moviesFromServer` in the component instead of assigning it to another const? 🤔

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #1754, opened 2023-12-07
Develop

jacekkosiba:
[DEMO LINK](https://jacekkosiba.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,21 +4,20 @@ import {
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+import searchTextInString from './searchTextInString';
 
 export const App: FC = () => {
   const [query, setQuery] = useState('');
   const [visibleMovies, setVisibleMovies] = useState(moviesFromServer);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,9 +1,30 @@
...
+  const handleChange: ChangeEventHandler<HTMLInputElement> = (event) => {
+    setQuery(event.target.value.trimStart());
+  };
+
+  useEffect(() => {
+    const filteredMovies = moviesFromServer.filter(movie => {
+      const isInTitle = searchTextInString(movie.title, query);
+      const isInDescription = searchTextInString(movie.description, query);
+
+      return (isInTitle || isInDescription) && movie;
```
zaura333, line 22: I'm not sure if checking for a movie is necessary, since it's an object of type Movie and this function is already using its properties (movie.title)

zaura333 (approved):
Looks great! I haven't stored visibleMovies in state and instead used basic variable, but that's a very interesting approach!

JoyCoffeeAddict (approved):

PR #1760, opened 2023-12-10
Solution

web3-sorcerer:
[DEMO LINK](https://web3-sorcerer.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,18 @@
-import React from 'react';
-import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+import './App.scss';
 
-export const App: React.FC = () => {
+export const App = () => {
+  const [query, setQuery] = useState('');
+
+  const filtered = moviesFromServer.filter(movie => {
+    const queryWords = query.trim().toLowerCase().split(/\s+/);
+
+    return queryWords.every(word => movie.title.toLowerCase().includes(word)
+    || movie.description.toLowerCase().includes(word));
+  });
+
   return (
     <div className="page">
       <div className="page-content">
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,18 @@
...
-import './App.scss';
+import { useState } from 'react';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+import './App.scss';
+
+export const App = () => {
+  const [query, setQuery] = useState('');
+
+  const filtered = moviesFromServer.filter(movie => {
```
Timi2424, line 9: -create a `visibleMovies` variable containing filtered movies;

Timi2424 (approved):
Works, looks good!

src/App.tsx
```diff
@@ -1,9 +1,18 @@
...
+    const queryWords = query.trim().toLowerCase().split(/\s+/);
+
+    return queryWords.every(word => movie.title.toLowerCase().includes(word)
+    || movie.description.toLowerCase().includes(word));
+  });
```
mpiskadlo, line 14: I'm sure you can find a simpler way to do it. I don't think`every` is the right 'tool' here - [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) ;)
It could be done using filter (just as you did) and a couple of includes (lowercasing and trimming as well)

mpiskadlo (changes requested):
Search doesn't work as expected
Screenshot 2023-12-18 at 16 21 24

mvjl000 (changes requested):
I think the code looks good now but it still doesn't work correctly. Have you re-deployed it?

mpiskadlo (changes requested):
Sorry, I made a request from wrong account.

web3-sorcerer:
If I'm not mistaken redeploy is not needed in case when changes are made, push is enough.

JoyCoffeeAddict (approved):

PR #1764, opened 2023-12-11
solution

adrian200002:
[DEMO LINK](https://adrian200002.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -9,10 +9,10 @@ export const App: React.FC = () => {
 
   const filteredMovies = moviesFromServer.filter(movie => movie.title
     .toLocaleLowerCase()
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -20,12 +30,13 @@ export const App: React.FC = () => {
                 id="search-query"
                 className="input"
                 placeholder="Type search word"
+                onChange={(event) => setQuery(event.target.value)}
```
mvjl000, line 33: Create a handler function for this

mvjl000 (approved):
I would like to see a handler function for the input onChange, but I don't think it is worth requesting changes. Please keep that in mind in future!

Good job overall ;)

PR #1774, opened 2024-03-11
first solution. Can't deploy:

spencerspe:
[DEMO LINK](https://spencerspe.github.io/react_movies-list-filter/)
Can't deploy: keep getting this error:    npx browserslist@latest --update-db
  Why you should do it regularly: https://github.com/browserslist/browserslist#browsers-data-updating 
I tried solutions but none worked


lerastarynets (changes requested):
If you can't deploy, pls don’t ask for help in github, because this way mentors are only able to see your questions once the review is requested. Instead, pls always ask for help in **fe_chat** if you don't understand something so mentors can quickly help you

PR #1775, opened 2024-03-15
add task solution

AgnieszkaWw:
[DEMO LINK](https://AgnieszkaWw.github.io/react_movies-list-filter/) and add it to the PR description.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,42 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+interface Movie {
+  title: string;
+  description: string;
+  imgUrl: string;
+  imdbUrl: string;
+  imdbId: string;
+}
+
+const movies: Movie[] = moviesFromServer;
+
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,15 +1,42 @@
...
+  title: string;
+  description: string;
+  imgUrl: string;
+  imdbUrl: string;
+  imdbId: string;
+}
+
+const movies: Movie[] = moviesFromServer;
+
+const filteredMovies = (allMovies: Movie[], query: string): Movie[] => {
```
yevhenii-pyl, line 16: A small adjustment with naming for this helper function may help here.
It's a function, so it's suppose to do something, thus, instead of naming it `filteredMovies` I'd prefer `filterMovies` naming. 

It will just make more sense inside your component: `const filteredMovies = filterMovies(...)`, don't you think so?

yevhenii-pyl (approved):
Good job! I love your approach with extracting the helper function outside the component: as your code grows you definitely will want to create a separate `helpers.ts` file for that type of functions. 

However, don't forget to always think twice about naming. Right naming takes more time, but it always pays off: sometimes that's the only reason why something works wrong and is really hard to debug.

PR #1783, opened 2024-04-04
solution

weswojciech:
https://weswojciech.github.io/react_movies-list-filter/


src/App.tsx
```diff
@@ -1,9 +1,29 @@
...
+      movie.description.toLocaleLowerCase().includes(lowerCaseQuery),
+  );
+}
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleSearch: ChangeEventHandler<HTMLInputElement> = e => {
+    setQuery(e.target.value.trim());
+  };
```
nataliaklonowska, line 25: Don't use one-letter variable naming. (It could be confusing for another developer, who reads your code. Also, it would really hard to find this variable in the code editor using search) [[checklist]](https://github.com/mate-academy/react_movies-list-filter/blob/master/checklist.md)

nataliaklonowska (changes requested):

choeqq (approved):

PR #1801, opened 2024-05-12
Solution

przwojwwp:
[DEMO LINK](https://przwojwwp.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,8 +2,21 @@ import React from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+import { useState } from 'react';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -2,8 +2,21 @@ import React from 'react';
...
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const handleChange: React.ChangeEventHandler<HTMLInputElement> = event => {
+    setQuery(event.target.value);
+  };
+
+  const visibleMovies = moviesFromServer.filter(
+    movie =>
+      movie.title.toLowerCase().includes(query.trim().toLowerCase()) ||
```
yevhenii-pyl, line 16: `query.trim().toLowerCase()` I'd prefer creating a separate variable for it, to avoid repeating

yevhenii-pyl (approved):
Nice and clean! 👍 

Just a small adjustment I've noted may make code even cleaner, but can't see it as a reason to block the PR.

PR #1809, opened 2024-07-06
Solution

kyrylomanko:
[DEMO LINK](https://kyrylomanko.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,28 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [visibleMovies, setVisibleMovies] = useState(moviesFromServer);
+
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,9 +1,28 @@
...
+    const filtered = moviesFromServer.filter(
+      movie =>
+        movie.title.toLowerCase().includes(userInput.toLowerCase().trim()) ||
+        movie.description
+          .toLowerCase()
+          .includes(userInput.toLowerCase().trim()),
+    );
+
+    setVisibleMovies(filtered);
+  };
```
etojeDenys, line 24:
```suggestion

  const handleInput: React.ChangeEventHandler<HTMLInputElement> = event => {
    const userInput = event.target.value;

    setQuery(userInput); 
  };
  
    const visibleMovies = moviesFromServer.filter(
      movie =>
        movie.title.toLowerCase().includes(query.toLowerCase().trim()) ||
        movie.description
          .toLowerCase()
          .includes(query.toLowerCase().trim()),
    );
```
you don't need that state at all

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #1811, opened 2024-07-14
solution

Kozubowicz:
DEMO LINK](https://Kozubowicz.github.io/react_movies-list-filter/)


src/App.tsx
```diff
@@ -1,9 +1,39 @@
...
+  imgUrl: string;
+  imdbUrl: string;
+  imdbId: string;
+};
+
 export const App: React.FC = () => {
+  const [query, setQuery] = useState<string>('');
+
+  const filterMovies = (list: Movie[]) => {
+    let tmp = [...list];
```
natalia-klonowska, line 18: [checklist](https://github.com/mate-academy/react_movies-list-filter/blob/master/checklist.md): you don't need to create copy because filter is non-mutating method so it will return altered copy of moviesArray

src/App.tsx
```diff
@@ -1,9 +1,39 @@
...
+
+    if (query.length > 0) {
+      tmp = tmp.filter(
+        movie =>
+          movie.title
+            .toLocaleLowerCase()
+            .includes(query.toLocaleLowerCase().trim()) ||
+          movie.description
+            .toLocaleLowerCase()
+            .includes(query.toLocaleLowerCase().trim()),
```
natalia-klonowska, line 28: [checklist](https://github.com/mate-academy/react_movies-list-filter/blob/master/checklist.md): Don't repeat yourself. If you see that you copy/paste some block code, maybe, it's time to create a separate variable/function

src/App.tsx
```diff
@@ -1,9 +1,39 @@
...
+      );
+    }
+
+    return tmp;
```
natalia-klonowska, line 32: you don't need to create tmp but simply return result of filter method

natalia-klonowska (changes requested):

natalia-klonowska (approved):
Nice work 😃

PR #1815, opened 2024-09-21
Develop

przwojwwp:
[DEMO LINK](https://przwojwwp.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,8 +6,10 @@ import moviesFromServer from './api/movies.json';
 export const App: React.FC = () => {
   const [query, setQuery] = useState('');
 
-  const visibleMovies = moviesFromServer.filter(movie =>
-    movie.title.toLowerCase().includes(query.toLowerCase()),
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,9 +1,17 @@
...
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+
+  const visibleMovies = moviesFromServer.filter(
+    movie =>
+      movie.title.toLowerCase().includes(query.toLowerCase().trim()) ||
+      movie.description.toLowerCase().includes(query.toLowerCase().trim()),
```
natalia-klonowska, line 12: don't repeat trimming query and instead save its result in variable

natalia-klonowska (changes requested):
make deploy so your demo would work

natalia-klonowska (approved):
Great work 🚀

PR #1824, opened 2025-01-29
Task solution

Dominik-Dudek-code:
- [DEMO LINK](https://Dominik-Dudek-code.github.io/react_movies-list-filter/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,38 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
 
+export interface Movie {
+  title: string;
+  description: string;
+  imgUrl: string;
+  imdbUrl: string;
+  imdbId: string;
+}
+
+export function filterMovies(movies: Movie[], inputValue: string) {
+  if (inputValue === '') {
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,9 +1,38 @@
...
+    return (
+      movie.title
+        .toLocaleLowerCase()
+        .includes(inputValue.trim().toLocaleLowerCase()) ||
+      movie.description
+        .toLocaleLowerCase()
+        .includes(inputValue.trim().toLocaleLowerCase())
+    );
+  });
+}
```
zibi-dev: Trim and lowercasing happens on every movie which is not efficient it also happens two times. I would do it like that this looks clearer and more efficient now.

```suggestion
  const trimmedLowercaseInput = inputValue.trim().toLocaleLowerCase()

  return movies.filter(movie => {
    const isTitleMatching =  movie.title.toLocaleLowerCase().includes(trimmedLowercaseInput)
    const isDescriptionMatching =   movie.description.toLocaleLowerCase().includes(trimmedLowercaseInput)

    return isTitleMatching ||  isDescriptionMatching;
  });
}
```

Zibi95 (changes requested):

natalia-klonowska (approved):

PR #1864, opened 2025-08-12
add task solution

isapq:
https://isapq.github.io/react_movies-list-filter/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,8 +2,21 @@ import React from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+import { useState } from 'react';
 
 export const App: React.FC = () => {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -2,8 +2,22 @@ import React from 'react';
...
 import { MoviesList } from './components/MoviesList';
 import moviesFromServer from './api/movies.json';
+import { useState } from 'react';
 
 export const App: React.FC = () => {
+  const [searchQuery, setSearchQuery] = useState<string>('');
+
+  /* eslint-disable */
+  const searchInList = (write: string) => {
+    const query = write.toUpperCase();
```
pedro-ruas, line 12: You should also use `trim` to remove extra spaces. This is causing some tests to fail.
Correct would be: `const query = write.toUpperCase().trim()`

pedro-ruas (changes requested):
Failing tests. I've pointed one fail origin

